PR #79, opened 2015-01-21
Making Validator safer and more testcases

fonsecas72:


Halleck45:
Thank's for this contribution. Improving tests case is always a nice idea :+1:

PR #85, opened 2015-02-02
Fixes one fatal error

baptistedeleplace:
see https://github.com/Halleck45/PhpMetrics/issues/73

### DIFF
--- a/src/Hal/Metrics/Mood/Abstractness/Abstractness.php
+++ b/src/Hal/Metrics/Mood/Abstractness/Abstractness.php
@@ -30,8 +30,10 @@ class Abstractness {
 
         foreach($results as $result) {
             $rOOP = $result->getOOP();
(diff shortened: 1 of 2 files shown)


baptistedeleplace:
Doesn't really solve the problem but at least it doesn't crash.
In the cases it crashed, it will now render a partial report ; 
In the cases it was stable, nothing changed.

Halleck45:
First version of PhpMetrics has two mode:
+ without OOP analysis (faster)
+ with OOP analysis

I'll remove the first mode as soon as possible. This will fix this issue (so I think that this PR will not be used).

baptistedeleplace:
I tried both mode, and it crashed.
These two little fixes aren't solving the reasons of crashing, but at least
is doesn't crash anymore.

Halleck45:
Ok, you're write, at least this patch avoid Fatal errors. I'll merge it

PR #87, opened 2015-02-11
Fix #86 add generated time of html report

### LINKED ISSUE #86: generate datetime
Hi,
I think it could be helpfull to have the datetime of the generate file (html) printed inside the page.

fonsecas72:
I've added something like this:
date_metrics

Hope you like it. I can put it elsewhere if you wish :)


Halleck45:
Nice idea. Thank's for the contribution !

PR #88, opened 2015-02-11
Improve syntax checker

fonsecas72:
This will add symfony/console to PhpMetrics which allow us to improve code readability and eventually some more improvements in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "symfony/console": "~2.4",
         "hoa/ruler": "~1.0",
         "symfony/config": "~2.4",
+        "symfony/process": "~2.3",
         "symfony/yaml": "~2.4",
         "ext-dom": "*"
     },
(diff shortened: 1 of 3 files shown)


Halleck45:
Hi,

I see the advantages, but I not sure that I want to use another library only for this feature : phpmetrics.phar is already very large ( >2.5Mo).

If you're ok, I'll close this PR :/

Thank's you your work !

fonsecas72:
Sure I agree. I have another improvements in mind.
I'll PR again as soon as I finish with them.
BTW, this is not related with this PR but, did you considered using PHP-Parser instead of your own  tokenizer?
(just asking)

Halleck45:
Yes, I would like to use PHPParser + [Hoa Compiler](http://hoa-project.net/En/Literature/Hack/Compiler.html), but that's too complex today :/

Probably for PhpMetrics v2

fonsecas72:
Sounds like a nice hobby :P
Sorry, I don't meant to be nagging but, is there any reason why you don't follow PSR-2 CS?

PR #89, opened 2015-02-11
Fixes #50 remove parent and self from relational graph

### LINKED ISSUE #50: parent treated as a class
"parent" is shown as a class in the relations graph : (tested on [Mondrian](https://github.com/Trismegiste/Mondrian) )

image

It's not a big deal but it's a little confusing. And I think perhaps it might affect some other metrics if PhpMetrics is seeing this call as a static call ?

Hope this helps :beer:

fonsecas72:


Halleck45:
Hi,

What do you think about returning the name of the parent instead of "null" ?

fonsecas72:
Is that going to be useful?
The parent class is already caught in another place so their relation is already available.

Halleck45:
I don't know yet

I've to read documentations, but I think that calling a mother class is a call like another. So returning it allows to:
+ count number of effective calls for each method (metric: "average calls by method")
+ build the graph of calls of the application
+ probably more...

fonsecas72:
Yes, you are very likely to be right.
Then we should add more tests .e.g to very the number of calls in a method.

Halleck45:
Hi @fonsecas72. Did you work on this feature ?

fonsecas72:
I might. I don't know how exactly do you want me to fix this. I've just implemented a quick fix.
You think it is good enough? Any suggestion?

Halleck45:
Hi,

for the moment, the quick fix is enough for me. Thanks !

PR #68, opened 2014-12-28
Fix typo in word maintainability

mre:

PR #94, opened 2015-02-16
Fix tests - Maintainability word

fonsecas72:
We should update all namespaces and class names as well.


Halleck45:
You're faster than me :)

Halleck45:
> We should update all namespaces and class names as well.

Yes, I've merged #94 too fast :/

fonsecas72:
;P

Halleck45:
:)

I've created issue #95

fonsecas72:
> Yes, I've merged #68 too fast :/
I think is better to do this is another issue anyway.

PR #100, opened 2015-02-21
Improvements to Travis

fonsecas72:
* Added PHP7
* Added Graphviz
* plus small things 

I've added graphviz because it seems to me that travis sometimes fails because of it. see https://travis-ci.org/Halleck45/PhpMetrics/jobs/50954456 besides there are other packages that do the same so I believe this is a good practice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: php
 
+sudo: false
+
+addons:
+  apt_packages:
+    - graphviz
+
 php:
     - 5.4
     - 5.5
@@ -8,9 +14,6 @@ php:
     - hhvm
 
 before_script:
-  - sudo apt-get -qq update > /dev/null
-  - sudo apt-get -qq install graphviz > /dev/null
-  - phpenv rehash > /dev/null
   - composer selfupdate --quiet
   - composer install --prefer-dist --no-interaction --no-progress
   - ./vendor/bin/phpunit
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,20 @@
 language: php
+
 php:
-  - 5.5
-  - 5.4
+    - 5.4
+    - 5.5
+    - 5.6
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,10 +1,20 @@
...
+    - 7.0
+    - hhvm
+
 before_script:
-  - wget http://getcomposer.org/composer.phar
-  - php composer.phar install --dev --prefer-dist
+  - sudo apt-get -qq update > /dev/null
+  - sudo apt-get -qq install graphviz > /dev/null
+  - phpenv rehash > /dev/null
+  - composer selfupdate --quiet
```
TomasVotruba: This has no really added value, Travis keeps composer up-to-date.

fonsecas72: Because Travis only update composer when they update PHP which according to the docs happens every 30-60 days. http://docs.travis-ci.com/user/languages/php/#Installing-Composer-packages

TomasVotruba: I don't get the benefit.

fonsecas72: Besides getting an updated composer (which is enough I think) it avoids an annoying warning.

.travis.yml
```diff
@@ -1,10 +1,23 @@
 language: php
+
+sudo: false
+
+addons:
+  apt_packages:
+    - graphviz
```
Halleck45: Can we declare this dependence as "optional"?

Halleck45:
Cool, excellent PR.

it would be best in the future to write functional tests in `\Hal\Application\Binaries`

.travis.yml
```diff
@@ -1,10 +1,23 @@
...
+
 php:
-  - 5.5
-  - 5.4
+    - 5.4
```
TomasVotruba: This is inconsistent with 2 spaces, that are used everywhere else.

PR #103, opened 2015-02-21
Fix color may be not initialized

fonsecas72:
In order to be safer we should add the default making a ```$color``` variable to always be initialized.

Scrutinizer-ci issue: 
> The variable $color does not seem to be defined for all execution paths leading up to this point.


Halleck45:
You're right. Thanks

PR #111, opened 2015-03-01
Class without parent should return null as parent class

UFOMelkor:
Actual a class without a parent class returns its namespace because of the usage of the `$nameResolver->resolve()`. This causes wrong dependencies in the [HtmlFormatter](https://github.com/Halleck45/PhpMetrics/blob/6ef089ab9179e766a426836cdff078ee0556373c/src/Hal/Application/Formater/Summary/Html.php#L96).


Halleck45:
The case of "parent" dependencies is indeed complex, but you're right, the Resolver should not return anything if class has no parent. Thanks !

PR #112, opened 2015-03-01
Always use fqcn for class names.

UFOMelkor:
Otherwise most classes would exists two times within the dependency graph: once with short name and once with fqcn.


Halleck45:
Exact ! Ouch, that's a big bug :/ . Thanks !

PR #113, opened 2015-03-01
Interfaces belong to a namespace.

UFOMelkor:
Actual all interfaces belong to the global namespace.


Halleck45:
Excellent PR, thanks @UFOMelkor !

PR #121, opened 2015-03-07
travis: use travis' phpunit

TomasVotruba:


Halleck45:
Why do not use Composer's PhpUnit ?

PR #122, opened 2015-03-07
composer: min PHP 5.4 added

TomasVotruba:
According to https://github.com/Halleck45/PhpMetrics/commit/23a7b9a49ca8824afc685e845fd4262cd89c3344, looks like min version is PHP 5.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         }
     },
     "require": {
+        "php": ">=5.4",
         "twig/twig": "~1.15",
         "symfony/console": "~2.4",
         "hoa/ruler": "~1.0",


fonsecas72:
So it seems, see https://github.com/Halleck45/PhpMetrics/issues/101

Halleck45:
No, PHP 5.4 is required only for building phar or running tests. PHPMetrics (`./bin/phpmetrics`) can be executed on PHP 5.3

TomasVotruba:
I see. Closing in favor of https://github.com/Halleck45/PhpMetrics/pull/125

PR #82, opened 2015-01-21
Updating Symfony Progress Helper to Progress Bar

shaunl:
This fixes fixes #63.

Issue #63, opened 2014-12-23
PHP Deprecated Notice re: Symfony ProgressBar


ryan-mahoney:
PHP Deprecated:  "Symfony\Component\Console\Helper\ProgressHelper" is deprecated since version 2.5 and will be removed in 3.0. Use "Symfony\Component\Console\Helper\ProgressBar" instead. in /var/www/project/vendor/opine/container/vendor/symfony/console/Symfony/Component/Console/Helper/ProgressHelper.php on line 123

PR #123, opened 2015-03-11
Changed the ProgressHelper by the ProgressBar

patxi1980:
because ProgressHelper it is deprecated in Symfony 2.5, and it will be deleted in 3.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea
 /vendor/
 composer.lock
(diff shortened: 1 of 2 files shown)


UFOMelkor:
Duplicate of #82

See also #63

Halleck45:
Thanks for the contribution, but, indeed, this patch is a duplicate of #82 and #63

PR #126, opened 2015-03-25
Fix abbr text in Explorer view header

halfer:
Thanks for your work on this!

Spotted a couple of spelling mistakes, fixed in this PR.


Halleck45:
Thanks for this PR ! Correct spelling is very important for the credibility of PhpMetrics :+1:

PR #133, opened 2015-05-19
Saving some memory for echo.

dseguy:


nicktacular:
See https://github.com/serbanghita/Mobile-Detect/pull/403

Halleck45:
Yes, `echo x,x` is few faster, but that's no relevant when we not repeat this echo less than thousand and thousand iterations.

PhpMetrics is slow (see [this issue](https://github.com/Halleck45/PhpMetrics/issues/13)), but I prefer avoid micro-optimisations for the moment...

I'll close your PR. In any case, thank's for your contributing !

Issue #136, opened 2015-06-17
Project title?


temp:
It would be nice to have a configurable project title in the report.

Halleck45:
Hi,

this is fixed by #140

PR #140, opened 2015-07-23
Adding Template Configuration Factory

rexem:
Added Template Configuration Factory class to allow passing custom settings to the HTML rendering class (Formater/Summary/Html).

This specific PR allows passing "template-title" option (either via "--template-title" or "template/title" via YML) which solves issue #136 (allows changing HTML summary page title)

*Usage:
CLI:
```
phpmetrics --report-html=report.html --template-title=Test src
```

YML:
```
myconfig:
[...]
    template:
        title: Test
[...]
```


Halleck45:
Merged. thanks, that's a good idea !

PR #144, opened 2015-08-21
composer: allow min PHP 5.4

TomasVotruba:


TomasVotruba:
Tests pass, scrutinizer fails internally

PR #159, opened 2015-12-01
Fix issue where large files would result in PHP warning

thorerik:
This should fix that issue, don't really know about code style though.


Halleck45:
Ok, you're right : `sprintf` was missing :/

Thanks !

Issue #160, opened 2015-12-06
Class member access on instantiation seems to be causing a bug


fonsecas72:
The following code is creating some issues
https://github.com/ircmaxell/Tuli/blob/master/lib/Tuli/Command/PrintCFG.php#L32
e.g.
```php
(new Foo)->bar().
```
This is a PHP 5.4 feature http://docs.php.net/manual/en/migration54.new-features.php

![metrics_screen](https://cloud.githubusercontent.com/assets/1098293/11615886/6b4300fa-9c64-11e5-934f-62a971e2171f.png)

PR #162, opened 2015-12-07
Php7

Halleck45:
## PHP 7
+ Anonymous classes parsed
+ get returned type of method
+ spaceship operator
+ null colaesce operator

## Misc
+ way to get visibility of a method
+ way to get state of method
+ `Searcher` improved
+ introduced new `ReflectedReturn` object
+ improved detection of returned values

## Command
+ flag `--ignore-errors` in order to ignore syntax errors (For example, if you parse PHP7 code with older version of PHP)

## Metrics
+ cyclomatic complexity takes care of anonymous classes
+ afferent coupling takes care of anonymous classes
+ report display number of anonymous classes

## Fixes
+ lot of bugs :)
+ #160


fonsecas72:
@Halleck45 hi, by checking this repo 
https://github.com/etsy/phan
 I believe that this sort of things are still causing some trouble 
https://github.com/etsy/phan/blob/master/src/Phan/Database/Model.php#L17
But I still haven't seen it closely though.

fonsecas72:
I also see this (`method' or`) 
metrics_screen
but I don't known if it is related with the previous one.

src/Hal/Component/OOP/Extractor/ClassExtractor.php
```diff
@@ -46,6 +47,14 @@ public function __construct(Searcher $searcher)
      */
     public function extract(&$n, TokenCollection $tokens)
     {
+        // is PHP7 ?
+        $previous = $tokens->get($n - 2);
+        if($previous && T_NEW === $previous->getType()) {
+            // anonymous class
+            $class = new ReflectedAnonymousClass($this->namespace, trim('class@anonymous'));
```
stof: why trimming a static string ?

src/Hal/Component/OOP/Resolver/TypeResolver.php
```diff
@@ -0,0 +1,100 @@
...
+
+        if('null' == $string) {
+            return self::TYPE_NULL;
+        }
+
+        if(preg_match('!(^\[|^array\()!', $string)) {
+            return self::TYPE_ARRAY;
+        }
+
+        if(preg_match('!(^\[|^array\()!', $string)) {
```
stof: this condition is the same than the previous one

src/Hal/Component/OOP/Resolver/TypeResolver.php
```diff
@@ -0,0 +1,100 @@
...
+            return self::TYPE_ARRAY;
+        }
+
+        if(preg_match('!^new\s*(.*)!', $string, $matches)) {
+            return trim(preg_replace('!^(new\s+)!', '', $cased));
```
stof: this is broken. ``return new TypeResolved($arg);`` will return the wrong stuff (probably ``TypeResolved($arg)`` in this case).

And this needs to support anonymous classes too:

```php
return new class implements Countable, Iterator {
    // ...
};
```

should be reported based on the interfaces implemented by the anonymous class.

If there are none, it should be reported as ``object`` (even though this may be less common)

Halleck45: Thanks @stof. I hope that is fixed with [this commit](https://github.com/Halleck45/PhpMetrics/commit/da7e0438461467ca3f981a29bbad8aa03d724957)

Halleck45: @fonsecas72 I'm checking it. Thanks !

src/Hal/Component/OOP/Resolver/TypeResolver.php
```diff
@@ -0,0 +1,101 @@
...
+        if('null' == $string) {
+            return self::TYPE_NULL;
+        }
+
+        if(preg_match('!(^\[|^array\()!', $string)) {
+            return self::TYPE_ARRAY;
+        }
+
+        if(preg_match('!^new\s+class\s+!', $string, $matches)) {
+            return self::TYPE_ANONYMOUS_CLASS;
```
stof: this looks weird to me. anonymous class is not a type.
You could get some real type info by reading the implemented interfaces, and otherwise you should use ``object`` (you know it is an object, but you don't know its class)

src/Hal/Component/OOP/Resolver/TypeResolver.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        if(preg_match('!^(new\s+)(.*?)([\(;].*|$)!', $cased, $matches)) {
```
stof: rather than ``.*``, you should match things without a ``(`` or ``:`` in it, to be sure that the regex does not match too far.

src/Hal/Component/OOP/Resolver/TypeResolver.php
```diff
@@ -0,0 +1,101 @@
...
+            return $matches[2];
```
stof: you need to trim it, as you can have spaces between the name and the parenthesis.

And the type being returned is not a FQCN here. Is it an issue ?

stof: btw, the phpdocumentor team has a library about resolving types. You might want to look at it

Halleck45: @stof that's an idea. But I don't want to get "object". My need is to get the "function" of the given string (if possible). In PhpMetrics, I think (but I'm not sure) that anonymous classes should not be considered as classical classes : they have impact on afferent coupling, but generally they are like "inner class", and their visibility is localized.

In fact, when we consider returned types, only classes or interfaces are used by PhpMetrics in calculations.

Halleck45: @fonsecas72 the parsing of repository https://github.com/etsy/phan looks good. All number / relations seem correct.

Can you confirm ?

fonsecas72: yeah, it seem :+1:

Halleck45:
I don't found lot of PHP7 code.

I check this PR on:

+ my own code
+ [codeliner/php-ddd-cargo-sample](https://github.com/codeliner/php-ddd-cargo-sample)
+ [etsy/phan](https://github.com/etsy/phan)
+ I don't found another PHP7 projects :(

Results seem coherent.

I think I'll merge today

fonsecas72:
@Halleck45 to me this is good enough to be merged. I still want to make some tests/experiments but I can do that from master branch.
:+1:

UFOMelkor:
@Halleck45 With our internal php 7 project everything looks good :+1:

Halleck45:
Cool ! 

I merge (I think that is the biggest PR merged in PhpMetrics)

PR #164, opened 2015-12-08
Fixes #163 - Some dependencies are not visible on Hierarchical graph

### LINKED ISSUE #163: Some dependencies are not visible on Hierarchical graph
Sample
If we have the next relations in report.relations.js.twig the graph hides the `\\1\\DUM` dependency because he confuses it with a package.
```js
   var relations = [
        {
            "name":"\\1\\DUM\\FOO",
            "size":3000,
            "relations":["\\2\\1\\BAR","\\1\\DUM"]},
        {
            "name":"\\1\\DUM\\FIN",
            "size":3000,
            "relations":[]},
        {
            "name":"\\2\\1\\BAR",
            "relations":[],
            "size":3000},
        {
            "name":"\\1\\DUM",
            "relations":[],
            "size":3000
        }
    ];
```

fonsecas72:
This feels like a hack fix, but I couldn't find another solution.

The Hierarchical Edge Bundling is not prepared for namespaces/packages which are also classes.

Making the class names a little different from namespaces (by adding a space) fixes the problem.

### DIFF
--- a/templates/html/summary/report-relations.js.twig
+++ b/templates/html/summary/report-relations.js.twig
@@ -93,6 +93,7 @@
             var map = {};
 
             function find(name, data) {
+                name = (data ? name + ' ' : name);
                 var node = map[name], i;
                 if (!node) {
                     node = map[name] = data || {name: name, children: []};


fonsecas72:
metrics_graph_fixed

Halleck45:
I'm not understand what is the problem. Can you give me an example that I can reproduce ?

fonsecas72:
@Halleck45 I was hoping that the issue #163 would answer it.
 You can easily see the issue if you create a folder with the following structure:
* \\namespace\\DUM\\FOO (which relates with BAR and DUM)
* \\whatever\\YIN\\BAR
* \\namespace\\DUM (this is a class but it has the same name as the namespace)

Then the graph will not show the DUM class.

Halleck45:
Ah yes, I did not saw the issue :/

Ok. That's a beautiful hack :) . I don't see another solution, so I merge it

Thanks !

PR #166, opened 2015-12-10
Removed unneeded code

fonsecas72:
While I was looking at your code my IDE found this weird things.
If I'm not missing anything I think he is right. But if not then tests should be failing.

### DIFF
--- a/src/Hal/Component/OOP/Extractor/MethodExtractor.php
+++ b/src/Hal/Component/OOP/Extractor/MethodExtractor.php
@@ -191,7 +191,7 @@ class MethodExtractor implements ExtractorInterface {
      *
      * @param ReflectedMethod $method
      * @param integer $n
-     * @param TokenCollection[] $tokens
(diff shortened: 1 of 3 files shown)


src/Hal/Component/OOP/Extractor/MethodExtractor.php
```diff
@@ -191,7 +191,7 @@ private function extractContent(ReflectedMethod $method, $n, TokenCollection $to
      *
      * @param ReflectedMethod $method
      * @param integer $n
-     * @param TokenCollection $tokens
+     * @param TokenCollection[] $tokens
```
Halleck45: The type here is not a "collection of TokenCollection". TokenCollection is an iterator.

fonsecas72: oh of course

Halleck45:
I admit that this PR made me a little scared :)

Thanks for your help !

PR #180, opened 2016-02-05
fix calculation of duration

### LINKED ISSUE #179: Wrong Output of estimated duration to read and understand
When generating a report with cli output, an estimated time is calculated which shows how long it would take to read and understand the code.

`321 files have been analyzed. Read and understand these 20404 lines of code will take around 00 hour(s), 01 minute(s) and 38 second(s).`

This is based on a sum of seconds which are printed with `gmdate()`. But this shows up to be faulty, because if you have time > 24 hours (86400 secs) the hour is still 00. using `gmdate()` is the wrong way to calculate a duration.

This works (credits to http://stackoverflow.com/a/3172665/1609286):

```php
<?php
    /**
     * Format time in text
     *
     * @param null|double $v
     * @return string
     */
    private function formatTime($v) {
        return sprintf('%s hour(s), %s minute(s) and %s second(s)'
            , floor($v / 3600)
            , ($v / 60) % 60
            , $v % 60
        );
    }
```

This produces the correct output:

`321 files have been analyzed. Read and understand these 20404 lines of code will take around 24 hour(s), 48 minute(s) and 38 second(s).`

skors:
this fixes #179 and changes calculation of seconds to hours, minutes and seconds

### DIFF
--- a/src/Hal/Application/Formater/Details/Cli.php
+++ b/src/Hal/Application/Formater/Details/Cli.php
@@ -177,9 +177,9 @@ class Cli implements FormaterInterface {
      */
     private function formatTime($v) {
         return sprintf('%s hour(s), %s minute(s) and %s second(s)'
-            , gmdate('H', $v)
-            , gmdate('m', $v)
-            , gmdate('s', $v)
+            , floor($v / 3600)
+            , ($v / 60) % 60
+            , $v % 60
         );
     }
 
--- a/src/Hal/Application/Formater/Details/Cli.php
+++ b/src/Hal/Application/Formater/Details/Cli.php
@@ -170,14 +170,15 @@ class Cli implements FormaterInterface {
     }
 
     /**
-     * Format time in text
+     * Format duration in seconds into text representation
      *
      * @param null|double $duration Duration in seconds
      * @return string
      */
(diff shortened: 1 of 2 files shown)


Halleck45:
Hi,

Well done! Perhaps this is an opportunity to add a test. Can you make the `formatTime()` method public, and add a unit test in `tests/Hal/Application/Formater/Details/CliTest.php` ?

skors:
@Halleck45 i do not think it is the right way just to set the method public and test it. I think it is protected because it is just used internally in the class, so we should test where it is used. Otherwise you producing a harder way to refactor this class in future.

skors:
@Halleck45 

I think testing the terminate Method correctly is the better way. Our goal is to test if some code produces the sum of n seconds in total, the written string should contain the equivalent of its time as string is the defined format. This will test the protected method `formatTime()` better i think. And we could test it with multiple values (negative, zero, positive, faulty values etc).

Nevertheless i will look into it soon. thanks for phpmetrics!

skors:
@Halleck45 please review, i've added a test for checking correct returned duration string

Halleck45:
Excellent ! Thanks a lot !

PR #185, opened 2016-02-18
Phar is pushed to bintray

Halleck45:
I think that downloading phar from Github is not a good idea:

+ no statistic
+ it's hard to download an older release
+ Github is not a download platform

I would like to host all files to bintray. If this PR is merged, I'll create a redirection on my server, so we'll be able to download phar at `http://download.phpmetrics.org/phpmetrics.phar`.

@fonsecas72 @UFOMelkor, what do you think ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 addons:
   apt_packages:
     - graphviz
+    - ruby2.1
 
 php:
     - 5.4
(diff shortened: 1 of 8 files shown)


fonsecas72:
totally agree, I'll review this.

.travis.yml
```diff
@@ -17,7 +18,24 @@ before_script:
   - composer selfupdate --quiet
   - composer install --prefer-dist --no-interaction --no-progress
   - ./vendor/bin/phpunit
-  - composer update --prefer-dist --no-interaction --no-progress
```
fonsecas72: why are you removing this?

.travis.yml
```diff
@@ -17,7 +18,24 @@ before_script:
...
 
 script:
-  - ./vendor/bin/phpunit
+  - make test
+
+after_script:
```
fonsecas72: Maybe this should be at "before_deploy" ?

artifacts/phar/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	@echo Copying sources
+	@mkdir -p /tmp/phpmetrics-build
+	@cp * -R /tmp/phpmetrics-build
+	@rm -Rf /tmp/phpmetrics-build/vendor /tmp/phpmetrics-build/composer.lock
+
+	@echo Releasing phar
+	@sed -i "s/<VERSION>/`semver tag`/g" /tmp/phpmetrics-build/artifacts/phar/build.php
+
+	@echo Installing dependencies
+	@cd /tmp/phpmetrics-build && composer install --no-dev --optimize-autoloader --prefer-dist
```
fonsecas72: I believe you can use `composer install --no-dev --optimize-autoloader --prefer-dist -d /tmp/phpmetrics-build`

artifacts/phar/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+	@echo Building phar
+	@cd /tmp/phpmetrics-build && php artifacts/phar/build.php
```
fonsecas72: `php /tmp/phpmetrics-buildartifacts/phar/build.php` would be readable IMO

.travis.yml
```diff
@@ -17,7 +18,24 @@ before_script:
...
+  - gem install semver
+  - make build
+
+
```
fonsecas72: extra break line

.travis.yml
```diff
@@ -5,6 +5,7 @@ sudo: false
 addons:
   apt_packages:
     - graphviz
+    - ruby2.1
```
fonsecas72: This is needed just for deploying right? Maybe there is a way of installing only when deploying ?

.travis.yml
```diff
@@ -17,7 +18,24 @@ before_script:
   - composer selfupdate --quiet
   - composer install --prefer-dist --no-interaction --no-progress
   - ./vendor/bin/phpunit
```
UFOMelkor: Is there a reason why phpunit is executed twice?

.travis.yml
UFOMelkor: `install` and `update` do the same when there is no lock file, don't they?

fonsecas72: Yes, I wasn't aware that the composer.lock was not there. 
Then the phpunit is begin executed twice as you said.

fonsecas72:
Another thing I would like to see is a warning message before the current download link is moved.

Halleck45:
I'll fix all your feedbacks

About current download link, I think we should keep the `build/phpmetrics.phar` file in the repository, to avoid 404

.travis.yml
Halleck45: No : deployment is triggered only on `master` branch. I think that if I use `before_deploy`, tests will not be launch for Pull requests.

artifacts/phar/Makefile
Halleck45: I fou refer to [this pr](https://github.com/phpmetrics/PhpMetrics/pull/12/files), you see that a bug will occur if `vendor/autoload.php` file exists in the **current directory**. 

To avoid unexpected bugs, I prefer to chdir to another directory.

.travis.yml
Halleck45: I changed the lines without paying attention.

.travis.yml
Halleck45: Maybe, but I don't find how to do :( . Do you have any idea ?

Halleck45:
I've changed some configuration. Feel free to merge when you think that all is OK :)

.travis.yml
fonsecas72: Tests are triggered on "script" aren't they?
Your `after_script` is just doing "make build" which is related with the deployment and not with PR evaluation. I'm I right?

fonsecas72: Sorry, I have commented without seeing your changes...
Is the `[ "${TRAVIS_PULL_REQUEST}" = "false" ]` condition really necessary? From what I have read in the Travis documentation the deploy only occurs for the master branch by default. It should not run on Pull Requests?

UFOMelkor: Yes, only master by default (see here: https://docs.travis-ci.com/user/deployment/bintray#Branch-to-deploy-from)

I might be wrong but [`after_script` is executed **after** `deploy`](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle), so IMHO the created phar will not be pushed to bintray.

fonsecas72: yeah, you may be right. Also we may have to add the `skip_cleanup: true` option so the artefact is not stashed, right?

UFOMelkor:
@Halleck45 Have you looked at https://github.com/phpmetrics/PhpMetrics/pull/185#discussion_r53608245 and https://github.com/phpmetrics/PhpMetrics/pull/185#discussion_r53609862 ?

.travis.yml
Halleck45: Ouch you're right.

I'll fix it :/

Halleck45:
@UFOMelkor no :( . I've opened new PR #206

Issue #189, opened 2016-03-02
svg graphs broken in v1.8.3


pavarnos:
installed latest version.
ran phpmetrics from the command line
overcame memory limit problem from #188 by doubling memory to 1024M.
Chrome reports javascript errors. Ditto in Firefox and Safari
<img width="851" alt="phpmetrics_report_-_by_jean-francois_lepine" src="https://cloud.githubusercontent.com/assets/589595/13476293/045d0748-e12c-11e5-8284-4dd23d1d4323.png">

line 1618 begins
```javascript
d3=function(){function n(n)
```

theofidry:
Same issue, besides it's worth noting that `phpmetrics -V` returns `1.0.0` regardless of your version...

theofidry:
Actually I tried to use an older version (1.7) and still see this issue...

GWillmann:
Temporary workaround for me was to add `<script src="http://d3js.org/d3.v3.min.js" charset="utf-8"></script>` in the generated html file header...

theofidry:
Thanks for the workaround @GWillmann, without it phpmetrics is unusable as I can't even explore the stats anymore.

mrVictorios:
Hello,

i have the same issue when i use the composer versions. when i use phar is all fine
![bildschirmfoto 2016-03-07 um 20 36 20](https://cloud.githubusercontent.com/assets/7589465/13580864/63cdc458-e4a4-11e5-8590-e75a412b455c.png)

UFOMelkor:
I cannot reproduce the error on my local machine :-/ (linux, php 7, phpmetrics 1.8.3 / master, opera/firefox). Perhaps you may provide some information about your setup?

When looking at the screenshot of @mrVictorios it seems that something replaced the `<<` after `(s&&c)` with a `\`.

@theofidry Wrong version output has been fixed in 0d95352. It will be available with the next release.

mrVictorios:
Setup:
osx, php 5.6, phpmetrics / master

Ive compare the report from phar and composer version,
![bildschirmfoto 2016-03-08 um 23 59 37](https://cloud.githubusercontent.com/assets/7589465/13619539/def0a5c0-e589-11e5-9fba-703fde405a2e.png)
![bildschirmfoto 2016-03-08 um 23 59 19](https://cloud.githubusercontent.com/assets/7589465/13619540/df1f8796-e589-11e5-931a-27c5b2896236.png)

[...]

PR #191, opened 2016-03-07
#189 svg graphs broken

corretge:
AS @GWillmann proposed, it works fine.


Halleck45:
Hi,

thanks for your PR !

I'am agree to use CDN instead injecting files with Twig.

But I prefer:
+ add a `--offline` option to preserver compatibility
+ include all JS and CSS files (and not only D3.js)

Can you change your PR in this way ?

Halleck45:
Done with #199 :)

PR #193, opened 2016-03-09
Fixed reading "symlinks" path parameter from config file

viruzzo:
The `path.symlinks` configuration directive is specified in the validator but not applied in the application, I added the missing line of code.


Halleck45:
Thanks ! :)

Issue #188, opened 2016-02-28
Allowed memory size of XXX bytes exhausted


infectormp:
Get some memory size error when scan code. I try it with different memory_limit.

Allowed memory size of 268435456 bytes exhausted (tried to allocate 528384 bytes) in phar:///usr/local/bin/phpmetrics/src/Hal/Component/Token/TokenCollection.php on line 118

Allowed memory size of 536870912 bytes exhausted (tried to allocate 262144 bytes) in phar:///usr/local/bin/phpmetrics/src/Hal/Component/Token/TokenCollectionCompatibility.php on line 56

Allowed memory size of 536870912 bytes exhausted (tried to allocate 32 bytes) in phar:///usr/local/bin/phpmetrics/src/Hal/Component/Token/TokenCollectionCompatibility.php on line 34

Want suggest add option to ignore files by size ```--max-size```

YuraLukashik:
@infectormp it's the very huge amount of memory, so i don't think the reason is a huge size of file. May be there is a bug with recursion or something like it.
Could you provide the file causing the failure?
Could you provide your full terminal output?

infectormp:
this is full output
98/184 [==============>-------------]  53%PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 32 bytes) in phar:///usr/local/bin/phpmetrics/src/Hal/Component/Token/TokenCollectionCompatibility.php on line 34

Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 32 bytes) in phar:///usr/local/bin/phpmetrics/src/Hal/Component/Token/TokenCollectionCompatibility.php on line 34

Look like tcpdf_5.0.002\fonts cause this error. But when i try check files one by one all work fine.

YuraLukashik:
@infectormp which php version do you use? try php 7.0.0, it might provide us some useful info

Halleck45:
PhpMetrics use the `token_get_all()` native PHP function to parse code. We cannot change the way this function parse code and use memory. But we can try to optimize the PhpMetrics's code...

Meanwhile, you can:

+ avoid to parse tcpdf files (tcpdf is a library, if you want to get metrics about your code, avoid to parse vendors and libraries)
+ increase your [memory_limit](http://php.net/manual/en/ini.core.php#ini.memory-limit).

infectormp:
@YuraLukashik i try PHP7 with same result. 
[...]

PR #208, opened 2016-03-23
pass TokenCollection instead of Tokenizer to Metrics

pavarnos:
for #188. Tokenizer was caching all the tokens for all files in the whole project. It didn't need to: each file is processed in turn and only the processed results are kept for analysis. Refactored to pass a TokenCollection to Metrics instead of the Tokenizer. Removed the Cache (no longer needed)


pavarnos:
to calculate peak memory usage, i added

```php
        $output->writeln(memory_get_peak_usage(true) . ' bytes used'); 
```
to the end of RunMetricsCommand.php
but did not commit that change because it did not seem useful for normal users...

src/Hal/Component/Cache/Cache.php
```diff
@@ -1,42 +0,0 @@
-<?php
```
Halleck45: Why did you removed this file ?

pavarnos: it was no longer used anywhere in the code base (i checked carefully). can easily put it back if you think it needed? I thought it is not a BC break because phpmetrics is an independent "app", not like a library with a public interface that needs a major version bump for a change like that... maybe i was wrong?

Halleck45:
Hum, I prefer 2 PR:

+ one for this change (TokenCollection instead of Tokenizer)
+ one for removing dead code

Thanks a lot !

src/Hal/Application/Command/Job/DoAnalyze.php
```diff
@@ -101,18 +100,18 @@ public function execute(ResultCollection $collection, ResultCollection $aggregat
 
         // tools
         $classMap = new ClassMap();
-        $tokenizer = new Tokenizer(new CacheMemory());
```
Halleck45: @pavarnos `Cache` is used at least here :wink:

Halleck45:
@pavarnos you made a really hard work :+1:. On my computer, PhpMetrics is around 20% faster :sunglasses: 

I'll merge as soon you will revert your changes about cache.

src/Hal/Application/Command/Job/DoAnalyze.php
pavarnos: yes... in the current master branch. But once this pull request is merged the code does not need it. (I searched through the whole code base: "CacheMemory" is not mentioned anywhere)

Halleck45:
Please be patient @pavarnos , I hope I'll merge tomorrow :)

PR #223, opened 2016-05-16
changing http assets to protocol autodetecting

bluemanos:
For a https connection the phpmetrics html report didnt show any charts because of asset attached with http protocol.
I also attach the js library through https://cdnjs.com/ site.


Halleck45:
Hi,

Can you provide some documentation ? What about blank (spaces) chars ?

Halleck45:
Nice, good idea :)

PR #226, opened 2016-05-27
Build the phar with a lower version of php to enhance compatibility

### LINKED ISSUE #219: PHP Parse error with phpmetrics.phar v1.10.0
Hi,

I've installed PhpMetrics version v1.10.0 as phar archive, with PHP 5.4.24 (cli) on Windows 10

When I run this command: 

```
php phpmetrics.phar folder
```

I get this output:

```
PHPMetrics by Jean-Fran├ºois L├®pine <https://twitter.com/Halleck45>

PHP Parse error:  syntax error, unexpected 'list' (T_LIST) in phar://phpmetrics.phar/vendor/symfony/config/Definition/ArrayNode.php on line 335

Parse error: syntax error, unexpected 'list' (T_LIST) in phar://phpmetrics.phar/vendor/symfony/config/Definition/ArrayNode.php on line 335
```

UFOMelkor:
Currently the phar is built with PHP 5.6. Therefore composer choses a Symfony version that is not compatible with PHP 5.4.

This PR changes the PHP version to 5.4 to ensure that the choosen dependencies are compatible with PHP 5.4. The tradeoff is that if the chosen version for PHP 5.4 of one dependency is not compatible with a greater PHP version, we will get the same problem (which could not be solved then).

The other approach would be to use a higher version of PHP (7.0 or 5.6) but run composer with the `--prefer-lowest`. This will always support the higher versions and the lowest possible PHP version that could be accomplished. (while writing this I convince myself that this is the better solution^^)

Thoughts?

Fix #219

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ deploy:
   key: ${BINTRAY_KEY}
   on:
     branch: master
-    php: '5.6'
+    php: '5.4'
   skip_cleanup: true
 
 env:


Halleck45:
Hi,

Maybe we can add a matrix for composer dependencies ? But how ?

Halleck45:
Can you make also this change on the `version2_ast` branch please ?

PR #229, opened 2016-06-20
Format to use less space column

Webysther:
I made some changes to show less column size, for using in travis, gitlab-ci and etc.

Before:

before

After:

after


Halleck45:
That looks a good idea. Thanks :)

PR #231, opened 2016-07-10
Fix deprecated warning on phpunit tests

20uf:
Hello @Halleck45 ,

I discovered impairments on unit testing.

Regards,


20uf:
Ha ... I have started tests on my global phpunit and not from the vendors, my version is more recent (5.6)

I close the PR

Issue #203, opened 2016-03-20
Missing images in “online” report version


UFOMelkor:
When loading the assets from the “cloud” the new version of [`nv.d3.js`](http://nvd3.org/assets/js/nv.d3.js) uses images (`images/grey-plus.png` / `images/grey-minus.png`) as open/close-buttons within the *Explore* section. Therefore there are many 404 requests and missing images.

Halleck45:
Hi,

Closing this issue will be hard... Do you have any idea ?

UFOMelkor:
The version of `nv.d3.js` we include is `1.1.11`, the current version is `1.8.2`. The `indentedTree` is not more active supported since [1.2.0](https://github.com/novus/nvd3/releases/tag/v1.2.0) and [fully removed](https://github.com/novus/nvd3/commit/530a786c534e144cde357d2c3829f24ca6656cdc) since v1.7.1.
Perhaps we should think about switching to another tree plugin?

bluemanos:
How can we help to fix the issue? What kind of requirements js (?) plugin should have to implement it on the Explore tab of PhpMetrics.

AJBrindley:
Why don't you encode the images in Base64 in the report, then there is no need for external images.

UFOMelkor:
Sounds like a cool idea. :+1:
IMHO we should think about switching to another tree plugin in the long term anyway.

UFOMelkor:
Fixed by #233

PR #233, opened 2016-07-13
fix + / - images for Explore view

MrSecure:
by including them as base64 data object.


UFOMelkor:
Fixes #203

Halleck45:
Cool! Thanks @MrSecure

PR #241, opened 2016-08-23
metric documentation: first experiment

pavarnos:
Work in progress to show you what i was thinking... no tests yet. They will come later if you approve the ideas...

Each metric has its own class which contains documentation (as you suggested). 

Each documentation class has
- a const ID which can be the symbolic name eg use LinesOfCode::ID instead of 'loc' throughout the code base. This will make it clear that 'loc' in one place is the same as all the other 'loc's. (and make them very easy to find in phpStorm and other IDEs).
- a short title to use as the title of a column in a table (for example)
- a short description to use as the abbr title
- a longer description where we describe what the metric is for, how to use it, what a good value is, what a bad value is (acceptable range)
- zero or more links to source documents, more definition, academic papers etc
- a formula which describes how it is calculated
- render*() functions which can turn it into html for display to the user. Of specific interest is AbstractInformation::renderFormulaToHtml() which will automatically convert metric names into links to those metrics in the documentation (what base url should we use here?).

So I've started documenting maybe 1/10th of the metrics. A lot of missing information (eg could not find info on Kan's Defects) but this can be filled in later once the whole structure is in place.

for me to do later
- finish creating documentation classes for all metrics (that are displayed to the user)
- replace strings eg 'loc' with their ID throughout the code base
- create an html page to display it all
- edit the other parts of the html report to link to the documentation or show abbr or title='' attributes where relevant.
- tests

need help with (when finished)
- better descriptions and links

### DIFF
--- a/src/Hal/Metric/Information/AbstractInformation.php
+++ b/src/Hal/Metric/Information/AbstractInformation.php
@@ -1,0 +1,128 @@
+<?php
+namespace Hal\Metric\Information;
+
+abstract class AbstractInformation
+{
+    const ID = 'undefined';
+    const LINK_TEMPLATE = '/information/metric#{metricID}';
+
+    /** @var string */
+    protected $name = '';
+
+    /** @var string */
+    protected $shortDescription = '';
+
+    /** @var string */
+    protected $longDescription = '';
+
+    /** @var string[] */
+    protected $links = [];
+
+    /** @var string if empty, this is a base metric that is not calculated from any other */
+    protected $formula = '';
+
+    /**
+     * unique symbolic identifier used in BagTrait to store values for this metric
+     * @return string
+     */
+    public function getID()
+    {
+        return static::ID;
+    }
+
+    /**
+     * short descriptive name
+     * @return string
+     */
+    public function getName()
+    {
+        return $this->name;
+    }
+
+    /**
+     * A one line description suitable for html title mouse over hints
+     * @return string plain text
+     */
+    public function getShortDescription()
+    {
+        return $this->shortDescription;
+    }
+
+    /**
+     * A longer description containing html
+     * @return string html including p tags
+     */
+    public function getLongDescription()
+    {
+        return $this->longDescription;
+    }
+
+    /**
+     * return a url to the documentation for this
+     * @param null $metricID
+     * @return string
+     */
+    public function getUrl($metricID = null)
+    {
+        if (empty($metricID)) {
+            $metricID = $this->getID();
+        }
+        return str_replace('{metricID}', self::LINK_TEMPLATE, $metricID);
+    }
+
+    /**
+     * return a formula in html with links to the other metrics that it is calculated from (if any)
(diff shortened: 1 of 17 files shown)


Halleck45:
Hi @pavarnos ,

Adding documenttion is a good idea :+1: 

However, why do not add it to the website ? Maybe it will be easier to have :

+ one class per metric (with `Id`,  `title`, `short description` and `link to website`)
+ a markdown file for each metric, in the [website repository](https://github.com/phpmetrics/website)

What do you think about this idea ?

Concerning metrics, I can write a first draft in french, then they will be translated by the community

pavarnos:
I'm not sure what you mean by "one class per metric"? You mean one PHP class? Per cluster of metrics (eg Halstead generates 15 metrics)

I did it in PHP because I wanted to make it available in the code in one central location. So you could use these classes to 
- generate your web site documentation maybe as part of your build process?
- generate a documentation page as part of the html report (so it is all one package: works offline)
- use the classes for column titles in CSV, html, json and other report formats (so you don't repeat yourself and can refer to the metric symbolically), and can generate tooltip definitions etc from a singe source.
- perhaps some other uses will occur to you once it is all in one place and callable (rather than scattered through the project in comments, documentation, code etc)

Just an idea. Happy to delete the pull request if it is going the wrong way for you. Let me know soon: was going to spend another couple of hours on it tonight...

pavarnos:
Random idea: what if longDescription and formula was replaced with a link to the full gh-pages documentation. Writing the longDescription in html inside php was going to be pretty awkward anyway...

src/Hal/Metric/Information/AbstractInformation.php
```diff
@@ -0,0 +1,128 @@
+<?php
```
Halleck45: This class looks really complex, and has many responsabilitites

+ each class is a value object and represents a cluster of metrics
+ but each class ahs also the responsability of rendering itself as string and HTML

I think that the methods `*Html()` should be moved in a renderer or directly used in the HTML templates

Halleck45:
> I'm not sure what you mean by "one class per metric"? You mean one PHP class? Per cluster of metrics (eg Halstead generates 15 metrics)

per cluster of metrics I think

> what if longDescription and formula was replaced with a link to the full gh-pages documentation. Writing the longDescription in html inside php was going to be pretty awkward anyway...

Yes, that looks a good idea

pavarnos:
is it worth continuing with this? i think maybe i misunderstood what direction you wanted to go in.
don't want to hold you up: v2 looks very good. can't wait to see it released. should i delete this PR?

Halleck45:
Hi,

> i think maybe i misunderstood what direction you wanted to go in

Neither do I :/

I don't think that one class by metric is the good solution : that's complex to group and links metrics. Maybe JSON files are preferable ? or directly markdown files ? I don't know

Yes, I'm really sorry but I think you can close your PR for the moment. Thanks for your help !!

PR #243, opened 2016-08-26
Add note about --offline to the README

halfer:
Let me know if another wording would be preferable.


Halleck45:
Hi,

that's perfect. Thanks :)

Issue #238, opened 2016-08-18
PhpMetrics v2.0.0 Fatal error on analyze folder (catchable error)


clagiordano:
My test environment:

``` bash
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux unstable (sid)
Release:    unstable
Codename:   sid
```

``` bash
$ php --version
PHP 7.0.8-3 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.8-3, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.4.0, Copyright (c) 2002-2016, by Derick Rethans
```

``` bash
$ ../PhpMetrics/bin/phpmetrics --version
PhpMetrics v2.0.0 <http://phpmetrics.org>
by Jean-François Lépine <https://twitter.com/Halleck45>
```

``` bash
$ ../PhpMetrics/bin/phpmetrics --report-html=build/metrics ./src
 27/34 [======================>-----]  79% 2 secsPHP Catchable fatal error:  Object of class PhpParser\Node\Expr\PropertyFetch could not be converted to string in PhpMetrics/src/functions.php on line 84
PHP Stack trace:
PHP   1. {main}() PhpMetrics/bin/phpmetrics:0
PHP   2. Hal\Application\Application->run() PhpMetrics/bin/phpmetrics:6
PHP   3. Hal\Application\Analyze->run() PhpMetrics/src/Hal/Application/Application.php:56
PHP   4. PhpParser\NodeTraverser->traverse() PhpMetrics/src/Hal/Application/Analyze.php:86
PHP   5. PhpParser\NodeTraverser->traverseArray() PhpMetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:64
PHP   6. PhpParser\NodeTraverser->traverseNode() PhpMetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:135
PHP   7. PhpParser\NodeTraverser->traverseArray() PhpMetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:84
PHP   8. Hal\Metric\Class_\Structural\LcomVisitor->leaveNode() PhpMetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:139
[...]

PR #247, opened 2016-11-14
v2: fix getNameOfNode when a node is a PropertyFetch

clagiordano:
see issue #238 for details


Halleck45:
Hi,

your PR is empty. I think that I've already done this patch on the `v2_ast` branch. Isn't ?

clagiordano:
Yes, we have made the same fix.

PR #257, opened 2017-02-16
Docs change: add symlink to run from anywhere

vaughany:
Added the `ln` command to symlink the composer-installed binary to /usr/local/bin/ so that it can be run from any location.


Halleck45:
Hi,

why do not suggest to add ̀`~/.composer/vendor` to the `$PATH` var ?

Halleck45:
I've just fixed documentation with #e2c5e531b14fcf34102cda7c3f015960a6c3834f

Issue #258, opened 2017-02-16
PhpMetrics 2.0 should use nikic/php-parser at least in version 3


niconoe-:
Hi,

I'm running PhpMetrics in version 1.10 and I'm trying to make it work with the new version 2.0.0.

I encounter a trouble with the parser that is unable to parse code from PHP 7.1.

Please, can you use the nikic/php-parser package in version 3.0.0 or above to make PHP 7.1 code parsable in the version 2.0.0 of PhpMetrics?

Thanks a lot.

Regards,

Halleck45:
Hi,

you're right. The [`composer.json`](https://github.com/phpmetrics/PhpMetrics/blob/master/composer.json#L37) now requires `v3.*̀

Thanks for your help !

PR #259, opened 2017-02-16
[Issue #258] Update composer.json to use the nikic/php-parser in version 3 to allow parsing PHP 7.1 sources.

niconoe-:


Halleck45:
Hi,

thanks for your help. This issue has already been patched

PR #260, opened 2017-02-16
Fix undefined MaintainabilityIndex on interfaces.

niconoe-:
As InterfaceMetric extends ClassMetric, during the consolidation, both parsed interfaces an classes where added into the list of `$classes`.
This resulted in an array `$classes` containing all classes and interfaces, while this array was used into tables and graphs in the HTML rendering. As interfaces have no metrics calculated from, the rendering goes to fail, espacially the metric "MaintainabilityIndex" was missing.

Using the workaround of comparing the FQN of the metric classes used for each `$item`, we can avoid this confusion and split correctly interfaces data and class data in the consolidated result object.

### DIFF
--- a/src/Hal/Metric/Consolidated.php
+++ b/src/Hal/Metric/Consolidated.php
@@ -43,20 +43,17 @@ class Consolidated
         $project = [];
         $nbInterfaces = 0;
         foreach ($metrics->all() as $key => $item) {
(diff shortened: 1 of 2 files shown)


Halleck45:
Cool ! Thanks for your help :clap:

PR #264, opened 2017-03-05
Run on Trusty Beta Container

photodude:
- add PHP 7.1
- Runs the tests on the newer Trusty containers since this repo is not affected by the current mysql issues in the trusty beta container.

Added benefit, This provides the current HHVM version (3.18.1 as of this PR) and the correct hhvm nightly (current dev) These will also track with each release (i.e. will be 3.19 when 3.19 is released) and nightly will actually be the nightly dev release. ... **you are currently on EOL versions of HHVM since precise was dropped back at hhvm 3.6**

If testing against HHVM LST versions is desired follow this guide. https://docs.travis-ci.com/user/languages/php#HHVM-versions

NOTE: If testing against HHVM in php 7 mode is desired: HHVM php 7 mode currently has an issue that messes up composer under php 7 mode, see facebook/hhvm#7198 once that BUG is fixed HHVM PHP 7 mode testing could be added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,15 +19,17 @@ matrix:
   - php: hhvm-nightly
 
 before_script:
-- composer selfupdate --quiet
-- composer install --prefer-dist --no-interaction --no-progress
+  - composer selfupdate --quiet
+  - composer install --prefer-dist --no-interaction --no-progress
+  # Disable JIT compilation in hhvm, as the JIT is useless for short live scripts like tests.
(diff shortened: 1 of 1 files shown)


photodude:
I have not looked into or addressed the current known issues with PHP 7.0/7.1/nightly. Those should be considered separate from this PR

Halleck45:
Thank you for this PR ! Good job ! Can you remove the support of `PHP 5.4` of your the `.travis.yml` file please ?

photodude:
you are welcome

PR #270, opened 2017-04-03
project dependencies updated to work with php 7.1 and do not use newest phpunit

kazimierasbutkus:


Halleck45:
Hi,

yes you're right. This issue is now fixed

PR #272, opened 2017-04-11
added calculation for max cyclomatic complexity of class methods and use it for reporting and violations calculation

kazimierasbutkus:
CCN for method is more common practice than CCN of class and we see it as a good metric to find classes that have complicated methods that needs refactoring.


Halleck45:
Hi,

Excellent idea ! Thanks :clap:

PR #279, opened 2017-04-13
(Mostly) adjust PHPDoc types

UFOMelkor:
Some types in  PHPDoc are outdated.

Also: I added a `getName()` method to the interface `Hal\Metric\Metric` because every Metric has it and it is used in the class `Hal\Metric\Metrics`.

Perhaps you might have a look at [Hal\Component\Tree\Operator\CycleDetector:46](https://github.com/phpmetrics/PhpMetrics/blob/master/src/Hal/Component/Tree/Operator/CycleDetector.php#L46)? I'm not sure about this line.

### DIFF
--- a/src/Hal/Application/Analyze.php
+++ b/src/Hal/Application/Analyze.php
@@ -33,7 +33,7 @@ class Analyze
 {
 
     /**
-     * @var OutputInterface
+     * @var Output
      */
     private $output;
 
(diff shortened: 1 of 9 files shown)


Halleck45:
Thanks @UFOMelkor !

Halleck45:
> Perhaps you might have a look at Hal\Component\Tree\Operator\CycleDetector:46? I'm not sure about this line.

This line allow to avoid infinite loop

PR #288, opened 2017-06-19
Updated pmd priority levels

krukru:
PMD uses values 1 to 5 for assigning priority, 1 being the most urgent task and 5 least urgent.

> 1 - Change absolutely required. Behavior is critically broken/buggy.
2 - Change highly recommended. Behavior is quite likely to be broken/buggy.
3 - Change recommended. Behavior is confusing, perhaps buggy, and/or against standards/best practices.
4 - Change optional. Behavior is not likely to be buggy, but more just flies in the face of standards/style/good taste.
5 - Change highly optional. Nice to have, such as a consistent naming policy for package/class/fields...

See http://pmd.sourceforge.net/pmd-5.0.5/rule-guidelines.html

I think that changing the values here will have no other effect on the application, but will result in a correct pmd file.


UFOMelkor:
Could be a BC break …

krukru:
People who are using standard PMD processing tools are having issues with this. Info level messages have a priority of 0, which is undefined (jenkins reports them as critical) and critical issues (currently at priority 4) will be marked as info-level warnings.

People who noticed this and made custom adjustments will have BC broken, but all they need to do is drop their custom changes and use standardised PMD parsing tools

I don't think you need to worry about breaking BC here, since the feature was originally bugged - but it does need to be documented. I will update the PR with a changelog

krukru:
I couldn't find if you already are keeping track of your changes in a changelog, so I created one. I tried sticking to http://keepachangelog.com/en/1.0.0/ since it seems to be getting some good traction lately. I also populated the changelog with changes from previous releases (just for 2.x)

UFOMelkor:
LGTM :+1: 
Anyway we need to wait for @Halleck45 to have a look at it, because he is the only one with merge access.

krukru:
Hi there, it's been some time now. No commits from the maintainer. Is the project still alive?

theofidry:
Last commit is from April 12. It's not that old for an open source project. And although the tool is not perfect, it is relatively stable and does a decent job.

krukru:
While I do agree that the tool is stable at it's current state, the fact that there are 10 open pull requests without even a comment from the maintainer in the last 4 months raises concerns.

theofidry:
It raises valid concerns, but it's not dramatic easier. It's hard to be always very responsive for Open Source :) I do think having more maintainers would help though.

Halleck45:
Hi everybody,

we're back :) @UFOMelkor that's way my fault, I tought you've the `write` access on this repository :/

@krukru , can you remove the `changelog.md` file from your PR (that's the subject of another pull request)

krukru:
Removed the changelog. Do you want a separate PR for it or will you take care of the [changelog](https://raw.githubusercontent.com/phpmetrics/PhpMetrics/d811097e1b32a257400707a77214260eb2c57e0b/CHANGELOG.md)?

niconoe-:
I think changelog must be part of a new PR. I'm merging this one.
@Halleck45 I'll let you take care of the changelog for previous versions, and I can handle the changes between 2.2.0 and the future 2.3.0

PR #290, opened 2017-07-03
Change heading for LOC graph.

bangpound:
_Répartition_ in French would be better translated as _distribution_ in English. Percentile distribution
graphs are common visualizations for demographic data, but this graph is about lines of code


bangpound:
Maintainers: Thank you for your work on this really useful project!

Issue #280, opened 2017-04-13
`CyclomaticComplexityVisitor` fails with anonymous class


nyoro712:
Hi,

I've installed PhpMetrics 2.2.0 as phar archive.

When I analyzed a code contains an anonymous class, PhpMetrics failed.

Here is an example:

1. Create PHP code file named `test.php`

```php test.php
<?php
declare(strict_types=1);

namespace Foo;

class Bar
{
    public function getAnonymousClass()
    {
        return new class {};
    }
}

```
2. Run `phpmetrics --report-html="report" test.php` command in the directory where `test.php` exists.
3. I get this output:

```
[2K... 100% ...PHP Notice:  Undefined property: PhpParser\Node\Stmt\Class_::$namespacedName in phar://D:/php/phpmetrics/src/Hal/Metric/Class_/Complexity/CyclomaticComplexityVisitor.php on line 50

Notice: Undefined property: PhpParser\Node\Stmt\Class_::$namespacedName in phar://D:/php/phpmetrics/src/Hal/Metric/Class_/Complexity/CyclomaticComplexityVisitor.php on line 50
PHP Fatal error:  Uncaught Error: Call to a member function toString() on null in phar://D:/php/phpmetrics/src/Hal/Metric/Class_/Complexity/CyclomaticComplexityVisitor.php:50
Stack trace:
#0 phar://D:/php/phpmetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(127): Hal\Metric\Class_\Complexity\CyclomaticComplexityVisitor->leaveNode(Object(PhpParser\Node\Stmt\Class_))
#1 phar://D:/php/phpmetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(120): PhpParser\NodeTraverser->traverseNode(Object(PhpParser\Node\Expr\New_))
#2 phar://D:/php/phpmetrics/src/Hal/Component/Ast/NodeTraverser.php(60): PhpParser\NodeTraverser->traverseNode(Object(PhpParser\Node\Stmt\Return_))
#3 phar://D:/php/phpmetrics/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(101): Hal\Component\Ast\NodeTraverser->traverseArray(Array)
[...]

PR #299, opened 2017-09-01
Fix to support anonymous classes in PHP 7.1

FractalizeR:
Fix for issue #280. Anonymous classes are now supported by analysis.


tests/Metric/examples/cyclomatic_anon.php
```diff
@@ -0,0 +1,15 @@
...
+
+class C
+{
+    public function foo1()
+    {
+        return new class
+        {
+        };
+    }
+}
```
niconoe-: There is a missing EOL at the end of file here.

niconoe-:
I have accepted previous PR that make yours in conflict. Can you resolve them please?

Thanks a lot.

FractalizeR:
I've rebased my PR according to your merges. In my PR the code, that's generating the name moved into a static helper class. I think this is a little better, than have it being copied from class to class.

PR #302, opened 2017-09-14
Embed dependent files so generated reports can be used offline

dennisdegreef:
I'd like to be able to view these reports when being offline, instead of being dependent on third party providers of css and javascript.
This change allows this.


Halleck45:
Hi,

I've removed this feature from v2 in order to reduce the size of Phar archive.

Can we imagine another solution ?

dennisdegreef:
Hi,

Generating the phar file from a fresh master, results in the following size:
`3,6M phpmetrics.phar`

While applying my branch, it generates the phar file as followed:
`4,2M phpmetrics.phar`.

The difference for the phar file is currently 600kb.

The advantages are:
- Being able to render the results offline
- Being able to render the results historically, when previously rendered results may contain references to third party libraries that may not exist anymore at time of render.

Can we maybe compress the assets so the size will shrink?

Halleck45:
@niconoe- @fonsecas72 @UFOMelkor what do you think about this PR ?

@dennisdegreef can you rebase please ?

niconoe-:
IMO, if the dependent files are embeded instead of being download from external domain, it will present 2 advantages:
- Being able to run the generated site in offline mode
- Avoid overloading client network to fetch those files.

Except when files are hosted on specific configured CDN servers, it is better to host them in the sources directly.
Plus, today, it is only 600kb more, which is pretty low. I mean, I needed to add 13Mb in the docker image (in a PR) to install git for the option "--git", as a comparison...

But if we allow that, we must limit it to avoid having a huge archive file. I think we should before each release look at the phar size and find a way to reduce it when too big. But a final size of 4.2Mb today it not very big to me.

dennisdegreef:
Rebased :+1:

Issue #291, opened 2017-07-05
Incorrect Coupling counts


auroraeosrose:
Phpmetrics version 2.2.0 installed via composer

Calculations for efferent coupling in phpmetrics are counting identical (duplicate) classes as additional couplings
this is incorrect

Pdepends shows effluent coupling of 4 for the controllers\base class shown below

When dumping the calculated graph using $graph->asString() inside  Hal\Metric\System\Coupling I am getting this as a result

```

Controllers\Base -> Psr\Container\ContainerInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> ValidationException;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> ValidationException;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> Psr\Http\Message\ResponseInterface;
Controllers\Base -> InvalidArgumentException;
Controllers\Base -> InvalidArgumentException;
Controllers\Base -> InvalidArgumentException;

```

The phpmetrics output gives this as effluent coupling of 20 - which is utterly incorrect
The value should be 4

I'm not exactly sure what is failing that is allowing these duplicate nodes to be added - still digging through code

auroraeosrose:
Looks like the graph class is adding duplicate edges

There is no check to see if an identical Edge (one with the exact same to and from node) already exists in the graph

The correct fix would be to add a $this->hasEdge($to, $from) method and check it before adding an edge, or a check inside the addEdge() method in the Graph class to keep from getting duplicates

For now this is working as a horrid workaround - add at src/Hal/Metric/System/Coupling/Coupling.php#L44

```
$skip = false;
foreach($graph->getEdges() as $edge) {
[...]

PR #304, opened 2017-09-14
Bugfix: preventing duplicate edges from appearing in a graph when analyzing

FractalizeR:
This fixes a problem #291 and hopefully some others.

I've added a check if an edge already exists in a graph before adding it in analyses. I've decided to leave a possibility to have a graph with duplicate edges just in case (though such ones are very rare)

### DIFF
--- a/src/Hal/Component/Tree/GraphDeduplicated.php
+++ b/src/Hal/Component/Tree/GraphDeduplicated.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Hal\Component\Tree;
+
+class GraphDeduplicated extends Graph
+{
+    /**
+     * @var array list of already present edges in this graph
+     */
+    private $edgesMap = [];
+
+    /**
+     * @param Node $from
+     * @param Node $to
+     *
+     * @return $this
+     */
+    public function addEdge(Node $from, Node $to)
+    {
+        $key = $from->getUniqueId().'->'.$to->getUniqueId();
+
+        if (isset($this->edgesMap[$key])) {
+            return $this;
+        }
+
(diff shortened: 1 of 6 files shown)


Halleck45:
Hi,

First, thanks for your fix !

I'm not sure that's was my favorite solution...

+ each class must deal with `if()` conditions
+ `Graph` code is complexified.

What do you think about something like :

```php
 <?php

/*
 * (c) Jean-François Lépine <https://twitter.com/Halleck45>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Hal\Component\Tree;

/**
 * Class GraphDeduplicated
 *
 * Graph without duplicated edges
 *
 * @package Hal\Component\Tree
 */
class GraphDeduplicated extends Graph
{

    /**
     * @var array list of already present edges in this graph
     */
    private $edgesMap = [];

    /**
     * @param Node $from
     * @param Node $to
     * @return $this
     */
    public function addEdge(Node $from, Node $to)
    {
        $key = $from . '->' . $to;
        if (isset($this->edgesMap[ey])) {
            return $this;
        }

        $this->edgesMap[$key] = true;
        return parent::addEdge($from, $to);
    }
} 
``̀

So when classes (like `Coupling`) must work with uniques egdes, it can instanciate `GraphDeduplicated` instead of `Graph` ?

FractalizeR:
That's a great idea, I did so. The only problem is that `Node` currently doesn't have any `__toString()` method and this is not obvious to use `spl_object_hash()` as a string representation of the node. So I added `getUniqueId()` to Node class for our purpose. Is that ok?

PR #311, opened 2017-09-19
Fix bad preg_quote usage that fails the regexp and so, fails the Finder.

niconoe-:


Halleck45:
Hi @niconoe- ,

:+1: , good idea. But can you add unit test please ?

niconoe-:
And here they are :).

Using `preg_quote(implode('|', $this->excludedDirs))` makes fail the test.
Using the bugfixed `implode('|', array_map('preg_quote', $this->excludedDirs))` makes pass the test.

Regards,

PR #313, opened 2017-09-21
Fix release

niconoe-:


niconoe-:
Travis has faild because of a curl timeout... How can I re-run it without making a push?

PR #321, opened 2017-10-20
Adds www. in URLs phpmetrics.org

ottaviano:
For avoiding `ERR_ADDRESS_UNREACHABLE`.

### DIFF
--- a/src/Hal/Application/Application.php
+++ b/src/Hal/Application/Application.php
@@ -37,7 +37,7 @@ class Application
             }
 
             if ($config->has('version')) {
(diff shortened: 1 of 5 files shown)


niconoe-:
This is a good quickfix, but it seems the willing is to have a website reachable on http://phpmetrics.org and not http://www.phpmetrics.org.

See: https://github.com/phpmetrics/website/issues/3

@Halleck45 , what is you point of view of this?

Regards,

UFOMelkor:
Quickfix or not, it solves the problem for now :+1:

PR #337, opened 2017-12-23
Improve composer package version comparison

juliendufresne:
We assume that installed version should be compared to the latest
stable version of the package and not to beta, rc, ...
Hence, installed version in the next RC version (or beta or whatever)
should not be listed as old (color orangered)

### DIFF
--- a/src/Hal/Metric/System/Packages/Composer/Packagist.php
+++ b/src/Hal/Metric/System/Packages/Composer/Packagist.php
@@ -37,8 +37,10 @@ class Packagist
         // get latest version
         $latest = '0.0.0';
         foreach ((array)$json->package->versions as $version => $datas) {
-            $version = preg_replace('([^\.\d])', '', $version);
-            if (!preg_match('!\d+\.\d+\.\d+!', $version)) {
+            if ($version[0] === 'v') {
+                $version = substr($version, 1);
+            }
+            if (!preg_match('#^[\.\d]+$#', $version)) {
                 continue;
             }
             if ($compare = version_compare($version, $latest) == 1) {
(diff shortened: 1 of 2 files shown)


src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -37,8 +37,10 @@ public function get($package)
         // get latest version
         $latest = '0.0.0';
         foreach ((array)$json->package->versions as $version => $datas) {
-            $version = preg_replace('([^\.\d])', '', $version);
```
juliendufresne: The only non numeric character that should be accepted is the leading "v" character that is used by convention.
Some packages have changed from a non-leading "v" to a leading "v" format. This is why we need to remove it before running the `version_compare` function.

src/Hal/Report/Html/template/index.php
```diff
@@ -141,7 +141,7 @@
                                 return strcmp($a->name, $b->name);
                             });
                             foreach ($packages as $package) { ?>
-                                <tr<?php if (null !== $package->installed && $package->installed !== $package->latest) { echo ' style="color:orangered"'; }?>>
+                                <tr<?php if (null !== $package->installed && version_compare($package->installed, $package->latest) === -1) { echo ' style="color:orangered"'; }?>>
```
juliendufresne: If you use a package in beta, alpha, RC (i.e non stable) version of a package and there is no stable version above the one you installed, PhpMetrics should not warn you.

src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -37,8 +37,10 @@ public function get($package)
...
-            if (!preg_match('!\d+\.\d+\.\d+!', $version)) {
+            if ($version[0] === 'v') {
```
juliendufresne: The only non numeric character that should be accepted is the leading "v" character that is used by convention.
Some packages have changed from a non-leading "v" to a leading "v" format. This is why we need to remove it before running the `version_compare` function.

UFOMelkor:
Thank you @juliendufresne

PR #341, opened 2018-01-24
Simplify return

carusogabriel:


src/Hal/Component/Ast/NodeTraverser.php
```diff
@@ -27,11 +27,7 @@ public function __construct($cloneNodes = false, $stopCondition = null)
 
         if(null === $stopCondition) {
             $stopCondition = function($node) {
-                if($node instanceof Node\Stmt\Class_ || $node instanceof Node\Stmt\Interface_) {
-                    return false;
-                }
-
-                return true;
+                return !$node instanceof Node\Stmt\Class_ || !$node instanceof Node\Stmt\Interface_;
```
juliendufresne: I guess you meant `&&` instead of `||`

carusogabriel: Yes, reverted the logic, but didn't revert the operands :sweat_smile:

PR #347, opened 2018-05-07
Support nikic/php-parser:^4

### LINKED ISSUE #345: Update of nikic/php-parser dependency
Any upgrade plans of this dependency?

Seems like next versions  eg. psalm is using 4.x branch of that library, so basically I need to decide which one I'm going to keep.

Another solution would be phpmetrics-shim , which would only contain necessary .phar file 

What do you think?

UFOMelkor:
Resolves #345 

Upgrade-Guide for php-parser can be found [here](https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-4.0.md).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,23 @@ sudo: false
 dist: trusty
 language: php
 
+php:
+ - 5.5
+ - 5.6
+ - 7.0
+ - 7.1
+ - 7.2
+ - nightly
+ - hhvm
+ - hhvm-nightly
+
+env:
+  matrix:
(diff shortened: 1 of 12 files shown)


UFOMelkor:
@tarlepp Can you test this?

zdenekdrahos:
I confirm the same results with `nikic/php-parser` v3 and v4.
I've compared html reports and it looked fine.

screenshot from 2018-07-05 13 03 55

tarlepp:
@UFOMelkor oh sorry haven't had the time to test this :( but according to @zdenekdrahos test it seems to work as expected

PR #348, opened 2018-05-08
Exclude dev-versions from packagist

### LINKED ISSUE #342: Package version parsing can give wrong value
Hi,

With PhpMetrics v2.3.2 (probably true with previous versions), the package version parsing from Packagist can result in wrong version number.

Example, the package `liip/imagine-bundle` output the lastest version to be **`2.01.0`**, but the lastest release is **`1.9.1`**.

The issue is from the line https://github.com/phpmetrics/PhpMetrics/blob/master/src/Hal/Metric/System/Packages/Composer/Packagist.php#L40:
It strip all chars that are not a digit or a dot, but in the case of `liip/imagine-bundle`, a branch (which are display as a version in packagist) is named **`dev-upgrade-2.0-from-1.0`**, and so result into `2.01.0`.

A quick workaround if to ignore all versions that starts with `dev-` (which are all branch that does represent a version number)  
(And ignoing all versions that ends with `-dev` will get rid of _number branches_)

UFOMelkor:
Fixes #342

### DIFF
--- a/src/Hal/Metric/System/Packages/Composer/Packagist.php
+++ b/src/Hal/Metric/System/Packages/Composer/Packagist.php
@@ -37,6 +37,9 @@ class Packagist
         // get latest version
         $latest = '0.0.0';
         foreach ((array)$json->package->versions as $version => $datas) {
+            if (stripos($version, 'dev-') === 0) {
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


UFOMelkor:
Closing in favor of #337

PR #352, opened 2018-05-21
Add a calculate the maintainability index for all project

2ur1st:


UFOMelkor:
I'm really sorry that you have to wait and I would ask you to wait a little bit more, I will definitely have a look at this next days. Please ping me if I did not til monday.

2ur1st:
@UFOMelkor ping

UFOMelkor:
Sorry again for the delay, I couldn't get enought time to have a closer look at the Maintainability Index :-(

The Maintainability Index is a system metric that should be calculated for the project as a whole (of course it can be calculated for classes also).
Currently we calculate it for classes only, so adding it for the whole project is very desirable.

IMHO the formula for the metric is:
```
171 - 5.2 * ln(aveV) -0.23 * aveG -16.2 * ln(aveLOC) + 50 * sin(sqrt(2.4 * perCM))
```
Where average is the average calculated value per method.

In your PR you're using the `aveMI = (Σ (mi_1 … mi_n) / n)` which provides another result as calculating the MI for the complete system would.
Therefore I suggest we find another way to calculate the system wide Maintainability Index.

PR #355, opened 2018-06-06
Root path exclude conflict fix

fabianbadoi:
Hi,

Currently, trying to run phpmetrics on paths containing any default excluded names (vendor, test, spec) fails because no files are found.

Here are some examples:
```
/home/tester/project-root/main.php	path contains "test"
./projects/respect/folder/		path contains "spec"
```

### DIFF
--- a/src/Hal/Component/File/Finder.php
+++ b/src/Hal/Component/File/Finder.php
@@ -48,8 +48,8 @@ class Finder
     private $flags;
 
     /**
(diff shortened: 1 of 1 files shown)


UFOMelkor:
Looks good to me :+1: 
I've added a test case and merged it manually (eb9945b19d9322916c5b8de63b69fe34447ef474 and a658ac166e1ee308baea6106ad011f4062f737f2). Thank you very much :blush:

PR #356, opened 2018-06-13
fix for repeating bloc-ids

dumith-eranga:
Same bloc-id is generated across multiple classes in the violations report.


UFOMelkor:
Thank you :+1:

PR #376, opened 2019-07-16
Fix error in demo code

mmeyer2k:
The example should directly run the executable, not parse with PHP. This confused me for a a solid minute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ PhpMetrics provides metrics about PHP project and classes, with beautiful and re
 ## Quick start
 
     composer require phpmetrics/phpmetrics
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,4 +40,4 @@ See the LICENSE file.
 
 ## Contributing
 
-See the [CONTRIBUTING](doc/contributing.md) file.
\ No newline at end of file
+See the [CONTRIBUTING](doc/contributing.md) file.
```
niconoe-: I think the end of line here is incorrect, that's why you have a diff without a change.
I also think that it makes conflicts the PR. Could you please fix this and the PR will be applied very fast then?

Thanks a lot.
Regards,

PR #381, opened 2019-11-07
Fixing PSR-0 filename for Traverser

SunMar:
This fixes the following warning in the latest composer snapshot:

```
Deprecation Notice: Class Hal\Component\Ast\Traverser located in ./vendor/phpmetrics/phpmetrics/src/Hal/Component/Ast/NodeTraverser.php does not comply with psr-0 autoloading standard. It will not autoload anymore in Composer v1.11+.
```


gmponos:
Hello.. can you check if this fixes your issue?

https://github.com/phpmetrics/PhpMetrics/pull/384

gmponos:
Fixed with #384

PR #390, opened 2019-12-02
Remove superfluous docblocks that are adding no value

gmponos:


src/Hal/Component/Output/ProgressBar.php
```diff
@@ -9,16 +9,7 @@
...
 namespace Hal\Component\Output;
 
-    /**
-     * Class CliOutput
-     * @package Hal\Component\Output
-     */
-    /**
-     * Class CliOutput
-     * @package Hal\Component\Issue
-     */
```
gmponos: I am not sure if there was any reason why these were here?

gmponos:
I guess most of these were added by PhpStorm.

Issue #379, opened 2019-08-28
PhpMetrics v2.4.1 PHP Warning:  Division by zero


muhagunghidayat:
Hi,

Thank you for taking the time to report this bug :+1:

Run `phpmetrics --version` to get the version of PhpMetrics. 

**The latest version of PhpMetrics is v2.4.1**. Please try to update PhpMetrics and check if the bug persist before submitting new issue.
 
Here is a template for your issue. Please replace the words between braces with the right informations. 

---- 

Hi,

I've installed PhpMetrics {version} {with composer | as phar archive}.

PhpMetrics v2.4.1 <http://www.phpmetrics.org>
by Jean-François Lépine <https://twitter.com/Halleck45>

When I run this command:

```
php ./bin/phpmetrics --report-html=./myreport7 ./src/82files

```

I get this output:

```
Executing system analyzes...

PHP Warning:  Division by zero in /home/agung/phpmetrics/vendor/phpmetrics/phpmetrics/src/Hal/Metric/Consolidated.php on line 137
PHP Warning:  Division by zero in /home/agung/phpmetrics/vendor/phpmetrics/phpmetrics/src/Hal/Metric/Consolidated.php on line 138
PHP Warning:  Division by zero in /home/agung/phpmetrics/vendor/phpmetrics/phpmetrics/src/Hal/Metric/Consolidated.php on line 139
PHP Warning:  Division by zero in /home/agung/phpmetrics/vendor/phpmetrics/phpmetrics/src/Hal/Metric/Consolidated.php on line 140
PHP Warning:  Division by zero in /home/agung/phpmetrics/vendor/phpmetrics/phpmetrics/src/Hal/Metric/Consolidated.php on line 141
PHP Warning:  Division by zero in /home/agung/phpmetrics/vendor/phpmetrics/phpmetrics/src/Hal/Metric/Consolidated.php on line 142
LOC
    Lines of code                               0
    Logical lines of code                       0
    Comment lines of code                       0
    Average volume                              0
    Average comment weight                      0
    Average intelligent content                 0
    Logical lines of code by class              0
    Logical lines of code by method             0

Object oriented programming
    Classes                                     0
    Interface                                   0
    Methods                                     0
    Methods by class                            0
    Lack of cohesion of methods                 0
    
Coupling
    Average afferent coupling                   0
    Average efferent coupling                   0
    Average instability                         0
    Depth of Inheritance Tree                   0
    
Package
    Packages                                    0
[...]

PR #401, opened 2019-12-22
Fix warning "Division by zero" when no package is defined.

niconoe-:
Fixing #379 when no packages is declared in a project.

### DIFF
--- a/src/Hal/Metric/Consolidated.php
+++ b/src/Hal/Metric/Consolidated.php
@@ -119,22 +119,24 @@ class Consolidated
         $avg->outgoingCDep = 0;
         $avg->outgoingPDep = 0;
         $avg->classesPerPackage = 0;
-        foreach (array_keys($packages) as $eachName) {
-            /* @var $eachPackage PackageMetric */
-            $eachPackage = $metrics->get($eachName);
-            $avg->distance += $eachPackage->getDistance();
-            $avg->incomingCDep += count($eachPackage->getIncomingClassDependencies());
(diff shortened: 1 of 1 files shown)


src/Hal/Metric/Consolidated.php
```diff
@@ -119,22 +119,24 @@ public function __construct(Metrics $metrics)
...
-        foreach (array_keys($packages) as $eachName) {
-            /* @var $eachPackage PackageMetric */
-            $eachPackage = $metrics->get($eachName);
-            $avg->distance += $eachPackage->getDistance();
-            $avg->incomingCDep += count($eachPackage->getIncomingClassDependencies());
-            $avg->incomingPDep += count($eachPackage->getIncomingPackageDependencies());
-            $avg->outgoingCDep += count($eachPackage->getOutgoingClassDependencies());
-            $avg->outgoingPDep += count($eachPackage->getOutgoingPackageDependencies());
-            $avg->classesPerPackage += count($eachPackage->getClasses());
+        if ([] !== $packages) {
```
gmponos: would advice to use `count($packages) !== 0`

Reason is that this ensures type safety.. meaning that it ensures that `$packages` is an array..

niconoe-: In that case, I'll use `$sum->nbPackages !== 0` as this was already the result of the count.
Actually, because there's already `$sum->nbPackages = count($packages)` earlier in the same snippet, it is already sure that `$package` is an array. So can't we keep this condition?

If you prefer `$sum->nbPackages !== 0`, I'm ok too.

PR #403, opened 2019-12-22
Review Reporters architecture.

niconoe-:
Just code review on the reporters. I planned to add XML as new output format but it's a bit more complicated than expected. The bag of metrics is containing too much different stuff to do something generic for XML and it would requires to develop a dedicated reporter.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
     "require": {
         "php": ">=5.5",
         "ext-dom": "*",
+        "ext-json": "*",
         "ext-tokenizer": "*",
         "nikic/php-parser": "^3|^4"
     },
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,6 @@
     "require": {
         "php": ">=5.5",
         "ext-dom": "*",
-        "ext-json": "*",
         "ext-tokenizer": "*",
         "nikic/php-parser": "^3|^4"
     },
--- a/src/Hal/Report/Cli/Reporter.php
+++ b/src/Hal/Report/Cli/Reporter.php
@@ -4,19 +4,23 @@ namespace Hal\Report\Cli;
 use Hal\Application\Config\Config;
 use Hal\Component\Output\Output;
 use Hal\Metric\Consolidated;
+use Hal\Metric\Metric;
 use Hal\Metric\Metrics;
+use Hal\Report\ReporterInterface;
+use function array_slice;
+use function arsort;
+use function round;
+use function sprintf;
 
-class Reporter
+/**
+ * This class takes care about the global output into the STDOUT of consolidated metrics.
+ */
+final class Reporter implements ReporterInterface
 {
-
-    /**
-     * @var Config
-     */
+    /** @var Config */
     private $config;
 
-    /**
-     * @var Output
-     */
+    /** @var Output */
     private $output;
 
     /**
@@ -29,15 +33,36 @@ class Reporter
         $this->output = $output;
     }
 
-
+    /**
+     * {@inheritDoc}
+     */
     public function generate(Metrics $metrics)
     {
         if ($this->config->has('quiet')) {
             return;
         }
 
-        // grouping results
         $consolidated = new Consolidated($metrics);
+        $out = $this->reportConsolidated($metrics, $consolidated);
+
+        if ($this->config->has('git')) {
+            $out .= $this->reportGitUsages($consolidated);
+        }
+
+        if ($this->config->has('junit')) {
+            $out .= $this->reportUnitTesting($metrics->get('unitTesting'));
+        }
+
+        $this->output->write($out . "\n");
+    }
+
+    /**
+     * @param Metrics $metrics
+     * @param Consolidated $consolidated
+     * @return string
+     */
+    private function reportConsolidated(Metrics $metrics, Consolidated $consolidated)
+    {
         $sum = $consolidated->getSum();
         $avg = $consolidated->getAvg();
 
@@ -50,7 +75,16 @@ class Reporter
             $locByMethod = round($sum->lloc / $sum->nbMethods);
         }
 
-        $out = <<<EOT
+        $inheritanceTreeDepthReport = '';
+        $treeMetric = $metrics->get('tree');
+        if (null !== $treeMetric) {
+            $inheritanceTreeDepthReport = <<<EOT
+Depth of Inheritance Tree                   {$treeMetric->get('depthOfInheritanceTree')}
+
+EOT;
+        }
+
+        return <<<EOT
 LOC
     Lines of code                               {$sum->loc}
     Logical lines of code                       {$sum->lloc}
@@ -67,13 +101,13 @@ Object oriented programming
     Methods                                     {$sum->nbMethods}
     Methods by class                            {$methodsByClass}
     Lack of cohesion of methods                 {$avg->lcom}
-    
+
 Coupling
     Average afferent coupling                   {$avg->afferentCoupling}
     Average efferent coupling                   {$avg->efferentCoupling}
     Average instability                         {$avg->instability}
-    Depth of Inheritance Tree                   {$metrics->get('tree')->get('depthOfInheritanceTree')}
-    
+    {$inheritanceTreeDepthReport}
+
 Package
     Packages                                    {$sum->nbPackages}
     Average classes per package                 {$avg->classesPerPackage}
@@ -88,7 +122,7 @@ Complexity
     Average Weighted method count by class      {$avg->wmc}
     Average Relative system complexity          {$avg->relativeSystemComplexity}
     Average Difficulty                          {$avg->difficulty}
-    
+
 Bugs
     Average bugs by class                       {$avg->bugs}
     Average defects by class (Kan)              {$avg->kanDefect}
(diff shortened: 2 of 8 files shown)


src/Hal/Report/Cli/Reporter.php
```diff
@@ -4,19 +4,23 @@
...
+use Hal\Report\ReporterInterface;
+use function array_slice;
+use function arsort;
+use function round;
+use function sprintf;
+
+/**
+ * This class takes care about the global output into the STDOUT of consolidated metrics.
+ */
+final class Reporter implements ReporterInterface
```
gmponos: making it final wont it be a BC?

src/Hal/Report/Csv/Reporter.php
```diff
@@ -2,34 +2,39 @@
...
+use Hal\Report\ReporterInterface;
+use RuntimeException;
+use function array_map;
+use function dirname;
+use function fclose;
+use function file_exists;
+use function fopen;
+use function fputcsv;
+use function is_scalar;
+use function is_writable;
```
gmponos: Usage of use function is available from 5.6 and after but we support 5.5

src/Hal/Report/Cli/Reporter.php
```diff
@@ -4,19 +4,23 @@
 use Hal\Application\Config\Config;
 use Hal\Component\Output\Output;
 use Hal\Metric\Consolidated;
+use Hal\Metric\Metric;
 use Hal\Metric\Metrics;
-
-class Reporter
+use Hal\Report\ReporterInterface;
+use function array_slice;
```
Halleck45: Please do not use this syntaxe (incompatible with PHP 5.5)

src/Hal/Report/Cli/Reporter.php
Halleck45: I'm not in favor of final for OSS code -> we never know how code will be used

src/Hal/Report/Cli/Reporter.php
```diff
@@ -67,13 +101,13 @@ public function generate(Metrics $metrics)
...
     Lack of cohesion of methods                 {$avg->lcom}
-    
+
 Coupling
     Average afferent coupling                   {$avg->afferentCoupling}
     Average efferent coupling                   {$avg->efferentCoupling}
     Average instability                         {$avg->instability}
-    Depth of Inheritance Tree                   {$metrics->get('tree')->get('depthOfInheritanceTree')}
-    
+    {$inheritanceTreeDepthReport}
```
Halleck45: Hum, is there 2 empty lines when tree is not defined, and only 1 when defined ?

src/Hal/Report/Csv/Reporter.php
```diff
@@ -40,30 +45,27 @@ public function generate(Metrics $metrics)
...
+        $logDir = dirname($logFile);
+        if (!file_exists($logDir) || !is_writable($logDir)) {
+            throw new RuntimeException('You do not have permissions to write CSV report in ' . $logFile);
         }
 
-        $availables = (new Registry())->allForStructures();
-        $hwnd = fopen($logFile, 'w');
-        fputcsv($hwnd, $availables);
+        $availableMetrics = (new Registry())->allForStructures();
+        $logPointer = fopen($logFile, 'wb');
```
Halleck45: why `b` ? That's not a [standard flag for fopen](https://www.php.net/manual/fr/function.fopen.php)

src/Hal/Report/Csv/Reporter.php
```diff
@@ -40,30 +45,27 @@ public function generate(Metrics $metrics)
...
                 continue;
             }
-            $row = [];
-            foreach ($availables as $key) {
+            $row = array_map(static function ($key) use ($metric) {
                 $data = $metric->get($key);
-                if (is_array($data) || !is_scalar($data)) {
-                    $data = 'N/A';
-                }
+                return (!is_scalar($data)) ? 'N/A': $data;
```
Halleck45: why removing the condition `is_array` ? Why the objective of this refactoring ?

src/Hal/Report/Html/Reporter.php
```diff
@@ -5,18 +5,28 @@
 use Hal\Component\Output\Output;
 use Hal\Metric\Consolidated;
 use Hal\Metric\Metrics;
-
-class Reporter
+use Hal\Report\ReporterInterface;
+use function array_map;
```
Halleck45: please avoid the `use function` syntaxe

src/Hal/Report/Html/Reporter.php
```diff
@@ -5,18 +5,28 @@
...
+use function json_decode;
+use function json_encode;
+use function mkdir;
+use function recurse_copy;
+use const JSON_PRETTY_PRINT;
+
+/**
+ * This class takes care about the global report in HTML of consolidated metrics.
+ */
+final class Reporter implements ReporterInterface
```
Halleck45: please avoid final here -> we never know how OSS projects will be used

src/Hal/Report/Html/Reporter.php
```diff
@@ -29,47 +39,31 @@ public function __construct(Config $config, Output $output)
...
-        }
-        if (!file_exists($logDir . '/css')) {
-            mkdir($logDir . '/css', 0755, true);
-        }
-        if (!file_exists($logDir . '/images')) {
-            mkdir($logDir . '/images', 0755, true);
-        }
-        if (!file_exists($logDir . '/fonts')) {
-            mkdir($logDir . '/fonts', 0755, true);
+        $history = array_map(static function ($filename) {
```
Halleck45: why replacing all `foreach` by `array_map` ? I'm not sure that's more readable...

src/Hal/Report/Json/Reporter.php
```diff
@@ -2,33 +2,35 @@
 namespace Hal\Report\Json;
 
 use Hal\Application\Config\Config;
-use Hal\Component\Output\Output;
 use Hal\Metric\Metrics;
-
-class Reporter
+use Hal\Report\ReporterInterface;
+use RuntimeException;
+use function dirname;
```
Halleck45: same thing about `use function` syntax :)

src/Hal/Report/Json/Reporter.php
```diff
@@ -2,33 +2,35 @@
...
+use function file_exists;
+use function file_put_contents;
+use function is_writable;
+use function json_encode;
+use const JSON_PRETTY_PRINT;
+
+/**
+ * This class takes care about the global report in JSON of consolidated metrics.
+ */
+final class Reporter implements ReporterInterface
```
Halleck45: Same thing about final here

Halleck45: (and in all PR)

src/Hal/Report/Cli/Reporter.php
gmponos: It's weird that tests did not catch this btw... we must look at this..

src/Hal/Report/Csv/Reporter.php
niconoe-: I always use `b` for portability, as PHP itself recommend: 
```
For portability, it is strongly recommended that you always use the 'b' flag when opening files with fopen().
```

src/Hal/Report/Csv/Reporter.php
niconoe-: Well, `is_array()` and `!is_scalar()` are tautologies: if it is an array, it is definitly not a scalar, and if it is not an array, the refactored condition is the same than the previous one.

src/Hal/Report/Html/Reporter.php
niconoe-: IMO I prefer `array_map` over `foreach` for its ability to set typed arguments and typed return clearly in the code, while you can not do that with the foreach. Since I'm using PHP 7.1 or above, I started to naturally write my foreach loops as array_map for that reason.

I kow that's not applicatble to this project since it must supports PHP 5.5 too, but that's not the only advantage to me. 

When you must create an array looping on other one(s), using `foreach` will oblige you to firstly initialize your array, then your `foreach` will push data inside. But if it is a long loop, it can be very hard to read and find if your array to populate is not fed more than once, for example. With the `array_map`, even if you may have several `return` in your callback, you will only return 1 thing for 1 loop-process, and you can easily know what that thing is by directly spotting the `return` instructions. Also, using `array_map` will provide you directly the array you want to build instead of doing it in 2 statements: initialization, then appending.

I keep using `foreach` when I care about the keys of the array I want to create as `array_map` does not offer me that way of doing it (actually, it does, but then, it is clearly a disguised foreach...).

src/Hal/Report/Cli/Reporter.php
niconoe-: Removed `final` because of potential BC break.

src/Hal/Report/Html/Reporter.php
niconoe-: Removed

src/Hal/Report/Csv/Reporter.php
Halleck45: I didn't known. Ok :)

composer.json
```diff
@@ -33,6 +33,7 @@
     "require": {
         "php": ">=5.5",
         "ext-dom": "*",
+        "ext-json": "*",
```
Halleck45: ext-json is not required, but can be suggested => not required if you don't use the report

Halleck45:
@niconoe- what about this PR ? Could you update the composer file please ?

ecourtial:
@niconoe- This PR need to be rebased :)

composer.json
niconoe-: After more than 18 months, I finally removed it :laughing:

niconoe-:
PR finally rebased after 18 months or so... sorry I didn't take time before to do so.

It should be OK now.

niconoe-:
All reporters have been upgraded except HTML one in v3 (work in progress).

This PR no longer means something.

PR #404, opened 2019-12-22
Ensure the packagist license is an array so they can be displayed.

### LINKED ISSUE #340: PHP Warning showing in html output with no license for composer package
The license property can be empty when displaying information about installed composer packages in the html display index page - this will cause a warning to be thrown and displayed in the script because it assumes that a non-iterable value is legal to pass into foreach (it is not)

Probably could be fixed with a simple is_array check before the foreach call but I won't get a pr rolled yet today

```
Warning: Invalid argument supplied for foreach() in C:\Users\emsmith\AppData\Roaming\Composer\vendor\phpmetrics\phpmetrics\src\Hal\Report\Html\template\index.php on line 149 Call Stack:    0.0003     351568   1. {main}()
```

niconoe-:
Fixes #340

### DIFF
--- a/src/Hal/Metric/System/Packages/Composer/Packagist.php
+++ b/src/Hal/Metric/System/Packages/Composer/Packagist.php
@@ -1,6 +1,8 @@
 <?php
 namespace Hal\Metric\System\Packages\Composer;
 
+use function is_array;
+
 /**
  * @package Hal\Metric\System\Packages\Composer
  */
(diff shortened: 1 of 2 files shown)


src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -1,6 +1,8 @@
 <?php
 namespace Hal\Metric\System\Packages\Composer;
 
+use function is_array;
```
Halleck45: Please avoid too recent code. This syntax is not compatible witll all supported PHP version

src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -45,7 +47,7 @@ public function get($package)
                 $latest = $version;
                 $response->name = $package;
                 $response->latest = $version;
-                $response->license = $datas->license;
+                $response->license = is_array($datas->license) ? $datas->license : [$datas->license];
```
Halleck45: You will be compatible with any PHP version with:

```
$response->license = (array) $datas->license;
```

PR #406, opened 2019-12-26
Move templates out of src

gmponos:
I don't think that these should be included into src.

This will help for various tools like php-cs and if in the future we want to include a static analysis tool.


src/Hal/Component/Ast/Php5NodeTraverser.php
```diff
@@ -1,31 +0,0 @@
-<?php
```
niconoe-: Why removing this file? If PhpMetrics is executed using PHP 5, this class should be the one used as `NodeTravserver`. If you want to keep the compatibility with PHP 5, you have to keep this file IMO.

niconoe-: Travils is failing on PHP 5 beause of this:

```
1) Test\Hal\Component\Ast\NodeTraverserTest::testItCanBeInstantiated
Class 'Hal\Component\Ast\Php5NodeTraverser' not found
```

gmponos: pffffff !@#$...
my bad.. was testing various stuff and committed by mistake.. 

Good thing tests are there.

Issue #325, opened 2017-11-02
Formula for Maintainability index


krukru:
I noticed that the formula for maintainability index takes into consideration code comments. We frequently try to keep comments to a minimum and use self-explanatory code and php7 types whenever possible. Sadly this means we are almost aways 20 points down on the MI because of no phpdocs.

The formula you use is a small derivation of the original formula (see [reference](http://www.projectcodemeter.com/cost_estimation/help/GL_maintainability.htm)).

I would like to be able to provide a custom implementation for calculating MI, Microsoft Visual Studio formula looks very promising. Do you think that this can be done? I could help with a PR, just point me in the right direction.

niconoe-:
Hi!

The reference your are pointing seems to be a bit wrong according to my research.

The original formula written by Paul Oman and Jack Hagemeister in 1992 and promoted by the SEI in 1997 is the one with the percentage of the lines of comments.
If you get deeper in the formula, you may also see that the "magic" 2,4 constant is just an approximation of the real value (pi²/4), that's why you can sometimes find the "magic" constant to "2,46" instead of "2,4".

The formula without comment is a simpler formula in case of no usage of documentation in the metrics (foreign language from a fetch foreign project so docs are unreadable, very old and unconsistants docs, ...)

The formula with comments has a range [0;221[
The formula without comments has a range [0;171]

The Visual studio's formula is just the same as the formula without comments but spread on a 100% line. This means if the formula without comments says your class have an MI of 171, the MVS's formula will say your MI is 100. For a MI (without comments) of 85,5, MVS will say 50.

To me, there are only 2 formulas, but the one without comments is not really relevant if you have projects with real usefull documentation pieces in. Indeed, I'm not sure about implementing it into a second MI calculation, but if that's something you really need, I will let you make a PR for this. But it must be a new value that must not be taking into account in the MI graphic and some valuable reports.

BTW, using self-explanatory code with php7 types are good things to quickly understand the basics, but a PHPDoc or some inline comments for tricky business part of your project are really recommanded.

pattisahusiwa:
+1 for Maintainability Index Without comment.

It would be nice if we can add the option in console argument (configuration) instead of this

> An idea can be to add a checkbox "exclude comments for calculation of maintainability". If checked, graph use the MIwoc instead of the Mi. What do you thin about this proposition ?

niconoe-:
@Halleck45 As this is released in 2.6, do you think we can close that issue?

[...]

PR #407, opened 2019-12-27
Provide configuration option "--maintainability-index-without-comments".

niconoe-:
Gives new functionnalities mentionned into #325


src/Hal/Report/Html/template/js/graph-maintainability.js
```diff
@@ -37,31 +37,48 @@ function chartMaintainability() {
...
+            } else {
+                var text = '<strong>' + d.name + '</strong>'
+                    + "<br />Cyclomatic Complexity : " + d.ccn
+                    + "<br />Maintainability Index: " + d.mi;
+            }
             d3.select('.tooltip').html(text);
-            d3.select(".tooltip").style("opacity", 1);
+            d3.select(".tooltip")
+                .style("opacity", 1)
+                .style("z-index", 1);
```
niconoe-: Added "z-index" to fix issue moseover with the tooltip (tooltip was hidding the buubles).

src/Hal/Application/Config/Validator.php
```diff
@@ -56,20 +56,21 @@ public function help()
...
-    --version                           Display current version
+    --exclude=<directory>                       List of directories to exclude, separated by a comma (,)
+    --extensions=<php,inc>                      List of extensions to parse, separated by a comma (,)
+    --report-html=<directory>                   Folder where report HTML will be generated
+    --report-csv=<file>                         File where report CSV will be generated
+    --report-json=<file>                        File where report Json will be generated
+    --report-violations=<file>                  File where XML violations report will be generated
+    --git[=</path/to/git_binary>]               Perform analyses based on Git History (default binary path: "git")
+    --junit[=</path/to/junit.xml>]              Evaluates metrics according to JUnit logs
+    --maintainability-index-without-comments    Ignore comments when calculating Maintainability Index
```
Halleck45: I'm not in favor of adding arguments for metrics.

What do you think about :
- add a checkbox "do not care of code comments" behind the graph ? 
- add new metric "maintability index without comments" ?

?

gmponos:
> add new metric "maintability index without comments" ?
Ah.. I like this

niconoe-:
> I'm not in favor of adding arguments for metrics. Otherwise, we'll have tons of specific arguments, and tool will be harder to use.
> 
> What do you think about :
> 
> * add a checkbox "do not care of code comments" behind the graph ? When checked, chart is updatd with the other metric
> * add new metric "maintability index without comments" ?
> 
> ?

I'm not a fan of the checkbox onto the graph because this is not only about the bubble graph: I also changed the MI displayed in the badges next to the Google Page Rank table. A more global usage of this "checkbox" is required then, but it will just move the complexity of having lots of metric configuration into having too many information displayed in the report.

That's the reason I choose the idea of having a configuration argument. Of course it brings higher complexity in the phpMetrics application, but I thought it could bring simplicity to the end user by not overloding him/her with unwanted metrics values.

I could easily change this if you do not want to keep this as a configuration, but we should discuss about the format in that case. More globally, we should agree on how can we deal with new metrics, and new way of reporting some of them.


Finally, I don't understand your question about "add new metric "maintability index without comments" ?": this metric already exists, even if only displayed in the "all.html" page.

niconoe-:
In order to solve arguments, I decided to follow the advise of @Halleck45 and remove the option. I set a javascript (native) carousel in order to display the 2 bubble graphs (MI and MIwoC). I also change the Google Page rank panel to display both information in the badges.

Like I said, I was not a fan of the checkbox on the main page because I feel like it could reduce the UX. I think the carousel is better and all informations are accessible without clicking anywhere.

Is it an acceptable PR for you?
Thanks a lot.

Halleck45:
Thanks @niconoe- !!

PR #408, opened 2019-12-27
Provide new issue template format.

niconoe-:


gmponos:
Hello why not have it per organization..
check here for an example https://github.com/guzzle/.github

niconoe-:
> Hello why not have it per organization..
> check here for an example https://github.com/guzzle/.github

I don't know the good practices of OpenSources' community health files. I just noticed that the current templates are deprecated so I wanted to update them.
You look like to have better experiences as working on Guzzle which is largely more supported than PhpMetrics. If you think you can improve this PR, or doing it in another way, please do so: I really want to learn that kind of stuff ;)

Thank you!

PR #412, opened 2020-02-20
Add metrics description file

repzy:
Added file with list of descriptions of almost all metrics that available in csv report. Unfortunately not all metrics provided with deep description and links. But it's good starting point to representing of all available information provided by this library.


repzy:
@gmponos

Halleck45:
Thank's ! That is a good idea !

PR #420, opened 2020-03-28
Improved page rank report, and UI #416

### LINKED ISSUE #416: DX: Explain the Google Page Rank metric
first let me say, thanks for the great tool!

the home-screen of the report shows me classes sorted by google page rank

grafik

it would be great to have a few explaining words in this area. something along the lines:
- what is the class-rank is trying to tell me?
  - lower is better (?)
- what is a maintainability index?
  - highter is better (?)
- which facts are taken into consideration for this table - so, how to improve my scores - is it about class relations? properties? complexity?

I dont know what the table is trying to tell me, so the above suggestions might not be correct as far as the metric is concerned. please take them as a example

Halleck45:
This fixes #416 

* adds comment about page rank
* improves UI

Capture d’écran du 2020-03-28 17-55-58

### DIFF
--- a/templates/html_report/complexity.php
+++ b/templates/html_report/complexity.php
@@ -71,7 +71,7 @@ require __DIR__ . '/_header.php'; ?>
                     <?php
                     foreach ($classes as $class) { ?>
                         <tr>
-                            <td><?php echo $class['name']; ?></td>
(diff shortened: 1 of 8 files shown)


staabm:
Thx for working on it <3

PR #421, opened 2020-03-28
Improved UI with help and gradients

Halleck45:
* Adds some help messages
* minor improvements (number format)
* Adds gradient for metrics everywhere

Capture d’écran du 2020-03-28 23-26-46
overview

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -164,3 +164,53 @@ function getVersion()
 {
     return 'v2.6.0';
 }
+
+/**
+ * @param array $array
+ * @param string $attribute
+ * @param mixed $currentValue
+ * @return false|float
+ */
+function gradientAlphaFor($array, $attribute, $currentValue)
+{
+    // memory cache
+    static $caches;
+    if(null === $caches) {
+        $caches = [];
+    }
+
+    if(!isset($caches[$attribute])) {
+        // avoid to iterate over array too many times
+        $max = 0;
+        $min = 1;
+        foreach($array as $item) {
+            if(!isset($item[$attribute])) {
+                continue;
+            }
+
+            $max = max($max, $item[$attribute]);
+            $min = min($min, $item[$attribute]);
+        }
+
+        $caches[$attribute]['max'] = $max;
+        $caches[$attribute]['min'] = $min;
+    }
+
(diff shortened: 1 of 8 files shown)


src/functions.php
```diff
@@ -164,3 +164,54 @@ function getVersion()
 {
     return 'v2.6.0';
 }
+
+/**
+ * @param $array
+ * @param $attribute
+ * @param $currentValue
```
niconoe-: Missing types for arguments in the PHPDoc. Also, I don’t think this function should be able to return false.

src/functions.php
```diff
@@ -164,3 +164,54 @@ function getVersion()
...
+ * @return false|float
+ */
+function gradientAlphaFor($array, $attribute, $currentValue) {
+
+    // memory cache
+    static $caches;
+    if(null == $caches) {
```
niconoe-: Probably “===“ could be better

src/functions.php
```diff
@@ -164,3 +164,54 @@ function getVersion()
...
+        }
+
+        $caches[$attribute]['max'] = $max;
+        $caches[$attribute]['min'] = $min;
+    }
+
+    $max = $caches[$attribute]['max'];
+    $min = $caches[$attribute]['min'];
+
+
```
niconoe-: Too many empty lines here, you should remove one for readability.

src/functions.php
```diff
@@ -164,3 +164,54 @@ function getVersion()
...
+    $percent = (($currentValue - $min) * 100) / ($max - $min);
```
niconoe-: You should manage the case that min and max are equals, leading to division by 0

src/functions.php
```diff
@@ -164,3 +164,54 @@ function getVersion()
...
+    return round($percent / 100, 2);
+}
+
+/**
+ * Style a element according its position in range
```
niconoe-: “Style an element...”

src/functions.php
```diff
@@ -164,3 +164,54 @@ function getVersion()
...
+ *
+ * @param $array
+ * @param $attribute
+ * @param $currentValue
+ * @return string
```
niconoe-: Also missing types for arguments in PHPDoc

PR #424, opened 2020-04-01
Composer report, UI

### LINKED ISSUE #417: DX: Explain colors in "Composer dependencies"
when looking at "Composer dependencies", some dependencies have a different color then others

grafik

would be great the UI would explain to me why `firebase/php-jwt` is in red font? what does the UI try to tell me?

Halleck45:
Hi,

we need to merge #421 first

* Improves UI
* Add a new Composer page, with Github stars, downloads... to fix #417 

Capture d’écran du 2020-04-01 23-03-48
Capture d’écran du 2020-04-01 23-06-05

### DIFF
--- a/src/Hal/Metric/System/Packages/Composer/Packagist.php
+++ b/src/Hal/Metric/System/Packages/Composer/Packagist.php
@@ -17,9 +17,19 @@ class Packagist
         $response->latest = null;
         $response->license = [];
         $response->homepage = null;
+        $response->description = null;
         $response->time = null;
         $response->zip = null;
         $response->compare = null;
+        $response->type = 'unknown';
+        $response->github_stars = 0;
+        $response->github_watchers = 0;
+        $response->github_forks = 0;
+        $response->github_open_issues = 0;
+        $response->download_total = 0;
+        $response->download_monthly = 0;
+        $response->download_daily = 0;
+        $response->favers = 0;
 
         if (!preg_match('/\w+\/\w+/', $package)) {
             return $response;
@@ -31,6 +41,18 @@ class Packagist
         if (!isset($json->package) || !is_object($json->package)) {
             return $response;
         }
+
+        $response->type = $json->package->type;
+        $response->description = $json->package->description;
+        $response->type = $json->package->type;
+        $response->github_stars = $json->package->github_stars;
+        $response->github_watchers = $json->package->github_watchers;
+        $response->github_forks = $json->package->github_forks;
+        $response->github_open_issues = $json->package->github_open_issues;
+        $response->download_total = $json->package->downloads->total;
+        $response->download_monthly = $json->package->downloads->monthly;
+        $response->download_daily = $json->package->downloads->daily;
+        $response->favers = $json->package->favers;
 
         // get latest version
         $latest = '0.0.0';
--- a/templates/html_report/_footer.php
+++ b/templates/html_report/_footer.php
@@ -31,6 +31,12 @@
                         Violations (<?php echo $sum->violations->total;?>)
                     </a>
                 </li>
+                <li>
+                    <a href="composer.html">
+                        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="white" width="24px" height="24px"><path d="M0 0h24v24H0z" fill="none"/><path d="M3.9 12c0-1.71 1.39-3.1 3.1-3.1h4V7H7c-2.76 0-5 2.24-5 5s2.24 5 5 5h4v-1.9H7c-1.71 0-3.1-1.39-3.1-3.1zM8 13h8v-2H8v2zm9-6h-4v1.9h4c1.71 0 3.1 1.39 3.1 3.1s-1.39 3.1-3.1 3.1h-4V17h4c2.76 0 5-2.24 5-5s-2.24-5-5-5z"/></svg>
+                        Composer
+                    </a>
+                </li>
                 <?php if($config->has('junit')) { ?>
                     <li>
                         <a href="junit.html">
--- a/templates/html_report/composer.php
+++ b/templates/html_report/composer.php
@@ -1,0 +1,98 @@
+<?php
+require __DIR__ . '/_header.php';
+
+$packages = isset($project['composer']['packages']) ? $project['composer']['packages'] : [];
+usort($packages, function ($a, $b) {
+    return strcmp($a->name, $b->name);
+});
+if ([] === $packages) {
+    echo '<div class="row"><div class="column"><div class="bloc">No composer.json file found in this project</div></div></div>';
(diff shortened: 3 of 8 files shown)


src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -32,6 +43,18 @@ public function get($package)
...
+        $response->type = $json->package->type;
+        $response->description = $json->package->description;
+        $response->type = $json->package->type;
+        $response->github_stars = $json->package->github_stars;
+        $response->github_watchers = $json->package->github_watchers;
+        $response->github_forks = $json->package->github_forks;
+        $response->github_open_issues = $json->package->github_open_issues;
+        $response->download_total = $json->package->downloads->total;
+        $response->download_monthly = $json->package->downloads->monthly;
+        $response->downloads_daily = $json->package->downloads->daily;
```
niconoe-, line 55: You should probably be consistent between "download" and "downloads" for your `$response` attributes names.

src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -17,9 +17,20 @@ public function get($package)
         $response->latest = null;
         $response->license = [];
         $response->homepage = null;
+        $response->description = null;
         $response->time = null;
         $response->zip = null;
         $response->compare = null;
+        $response->type = 'unknown';
+        $response->description = '';
```
niconoe-, line 25: Was just set to `null` on line 20.

templates/html_report/_footer.php
```diff
@@ -31,6 +31,13 @@
                         Violations (<?php echo $sum->violations->total;?>)
                     </a>
                 </li>
+                <li>
+                    <a href="composer.html">
+
```
niconoe-, line 36:
```suggestion
```

templates/html_report/composer.php
```diff
@@ -0,0 +1,100 @@
+<?php
+require __DIR__ . '/_header.php'; ?>
+
+<?php
```
niconoe-, line 4:
```suggestion
require __DIR__ . '/_header.php';

```

templates/html_report/composer.php
```diff
@@ -0,0 +1,100 @@
...
+$packages = isset($project['composer']['packages']) ? $project['composer']['packages'] : [];
+usort($packages, function ($a, $b) {
+    return strcmp($a->name, $b->name);
+});
+if (empty($packages)) {
```
niconoe-, line 9: `$packages` is always an array, so you might consider using a strict equality instead:

 ```suggestion
if ($packages === []) {
```

File: templates/html_report/composer.php [hunk 6]
```diff
@@ -0,0 +1,100 @@
...
+    echo '<div class="row"><div class="column"><div class="bloc">No composer.json file found in this project</div></div></div>';
+    require __DIR__ . '/_footer.php';
+    return;
+}
+?>
+
+
```
niconoe-, line 16:
```suggestion
```

templates/html_report/composer.php
```diff
@@ -0,0 +1,100 @@
...
+<div class="row">
+    <div class="column">
+        <div class="bloc">
+            <h4><?php echo sizeof($packages); ?> Composer dependencies</h4>
```
niconoe-, line 20: `sizeof` is deprecated in favor of `count`

templates/html_report/composer.php
```diff
@@ -0,0 +1,100 @@
...
+                                    <span class="help-warning" style="position: absolute; right: 10px;">This package should be updated.</span>
+                                <?php } ?>
+                                <span class="badge"><?php echo $package->type; ?></span>
+                                <?php echo $package->description; ?>
+                            </div>
+                        </div>
+
+                        <div class="list-item-content">
+                            <?php $width = sprintf('%d%%', round(100 / 9, 0)); ?>
+                            <table class="table-metrics" style="=margin-left:50px;">
```
niconoe-, line 42:
```suggestion
                            <table class="table-metrics" style="margin-left:50px;">
```

Probably a typo here.

templates/html_report/index.php
```diff
@@ -40,33 +52,42 @@
...
                 </div>
-                <?php echo $this->getTrend('avg', 'bugs', true); ?>
             </div>
         </div>
     </div>
 
     <div class="row">
         <div class="column column-help">
-            <div class="bloc">
+            <div class="bloc"  style="min-height: 540px;">
```
niconoe-, line 90:
```suggestion
            <div class="bloc" style="min-height: 540px;">
```

templates/html_report/index.php
```diff
@@ -161,66 +182,28 @@
...
-
-        <div class="column">
-            <div class="bloc bloc-number">
-                <div class="label">Licences of Composer dependencies</div>
-                <?php if(0 === count($packages)) { ?>
-                    <div>No composer.json file found</div>
+                <?php
+                $packages = isset($project['composer']['packages']) ? $project['composer']['packages'] : [];
+                $packagesInstalled = isset($project['composer']['packages-installed']) ? $project['composer']['packages-installed'] : [];
+                if(0 === count($packages)) { ?>
```
niconoe-, line 191:
```suggestion
                if ([] === $packages) { ?>
```

templates/html_report/index.php
```diff
@@ -161,66 +182,28 @@
...
+                    <div class="help number-alternate"><div class="help-inner">No composer.json file found</div></div>
+                <?php } else {?>
+                    <div class="number">
+                        <?php echo sizeof($packages);?> dependencies
```
niconoe-, line 195:
```suggestion
                        <?php echo count($packages);?> dependencies
```

templates/html_report/composer.php
Halleck45: Hi,

I don't think that [sizeof](https://www.php.net/sizeof) is deprecated. Do you have any info ?

niconoe-: Sorry, my bad, I though it was deprecated in favor of count since it is an alias since long ago with no particular matter.
I found an RFC where sizeof was suggested as deprecated in PHP 7.2, but obviously, it was just a suggestion: https://wiki.php.net/rfc/deprecations_php_7_2#suggested_deprecations

Nevertheless, `sizeof` has different meanings accross all languages, while `count` is more oftenly considered as the native function used to count elements in an array, a set or a map. Also, I would prefer to use the expected function from PHP rather than an alias for it.

So, no, it is not deprecated, but I would suggest to prefer `count` :).

ecourtial:
Nice improvements.

templates/html_report/composer.php
Halleck45: No problem. I've updated code

PR #428, opened 2020-04-09
Way to analyze by layer

Halleck45:
#421 needs to be merged first

Capture d’écran du 2020-04-09 22-50-52
Capture d’écran du 2020-04-09 22-50-27

Does somebody can test this feature on his project ?


src/Hal/Report/Html/Reporter.php
```diff
@@ -105,10 +138,40 @@ public function generate(Metrics $metrics)
...
+            $this->renderPage($this->templateDir . '/html_report/all.php', $outDir . '/all.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/oop.php', $outDir . '/oop.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/complexity.php', $outDir . '/complexity.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/panel.php', $outDir . '/panel.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/violations.php', $outDir . '/violations.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/packages.php', $outDir . '/packages.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/package_relations.php', $outDir . '/package_relations.html', $consolidated, $history);
+            $this->renderPage($this->templateDir . '/html_report/composer.php', $outDir . '/composer.html', $consolidated, $history);
```
niconoe-, line 166: Do you think it could be a little be less verbose if you can
- Export the `$this->templateDir . '/html_report'` string outside the `foreach loop` ;
- Loop over all page name as each line are identical ?

src/Hal/Report/Html/Reporter.php
```diff
@@ -205,4 +268,12 @@ protected function getTrend($type, $key, $lowIsBetter = false, $highIsBetter = f
             $svg[$r]
         );
     }
+
+    /**
+     * @return bool
+     */
+    private function isHomePage()
+    {
+        return !$this->currentGroup;
```
niconoe-, line 277: As `currentGroup` is a string, it should be something like this:
 ```suggestion
        return '' === $this->currentGroup;
```

File: src/functions.php [hunk 3]
```diff
@@ -214,3 +214,16 @@ function gradientAlphaFor($array, $attribute, $currentValue)
 function gradientStyleFor($array, $attribute, $currentValue) {
     return sprintf(' style="background-color: hsla(203, 82%%, 76%%, %s);"', gradientAlphaFor($array, $attribute, $currentValue));
 }
+
+/**
+ * Calculate percentalies
+ *
+ * @param $arr
```
niconoe-, line 221: Missing typehit of `$arr`, is it `mixed[]`?

templates/html_report/_header.php
```diff
@@ -48,8 +48,21 @@
...
 
 <div class="page">
-    <div class="content">
+    <div class="content content-first">
         <div class="report-details">
             Created at <?php echo date('Y-m-d H:i:s'); ?>
         , with PHPMetrics <?php echo getVersion(); ?> (<a href="https://twitter.com/Halleck45">Jean-François Lépine</a>).
         </div>
+
+        <?php if([] !== $this->groups) {?>
```
niconoe-, line 57:
```suggestion
        <?php if ([] !== $this->groups) {?>
```

templates/html_report/_header.php
```diff
@@ -48,8 +48,21 @@
...
+        <div class="content-full group-tabs">
+            <ul class="tabs">
+                <li class="<?php if(!$this->currentGroup) { echo 'active'; }?>"><a href="<?php echo $this->assetPath;?>index.html">All</a></li>
```
niconoe-, line 60:
```suggestion
                <li class="<?php if ('' === $this->currentGroup) { echo 'active'; }?>"><a href="<?php echo $this->assetPath;?>index.html">All</a></li>
```

templates/html_report/_header.php
```diff
@@ -48,8 +48,21 @@
...
+                <?php foreach($this->groups as $group) { ?>
```
niconoe-, line 61:
```suggestion
                <?php foreach ($this->groups as $group) { ?>
```

templates/html_report/css/style.css
```diff
@@ -340,7 +345,33 @@ a[target="_blank"]::before {
     color: #333;
     text-align: left;
 }
-
+.bloc .number, .bloc .number-alternate {
```
niconoe-, line 348: Some colors definitions are not consistant between uppercase and lowercase (#ABCDEF vs #abcdef).

You should be consistant in your whole project, whatever you decide which case you want to apply, for readability.

templates/html_report/index.php
```diff
@@ -159,68 +180,32 @@
...
+                    <div class="label">
+                        <a href="composer.html">Composer</a>
                     </div>
-                </div>
-            </div>
-        </div>
+                    <?php
+                    $packages = isset($project['composer']['packages']) ? $project['composer']['packages'] : [];
+                    $packagesInstalled = isset($project['composer']['packages-installed']) ? $project['composer']['packages-installed'] : [];
+                    if([] === $packages) { ?>
```
niconoe-, line 192:
```suggestion
                    if ([] === $packages) { ?>
```
But this probably comes from the #424. If there is a dependency with a rebase required, please let it go.

README.md
```diff
@@ -31,6 +31,34 @@ Then open the generated `./myreport/index.html` file in your browser.
...
+        },
+        {
+            "name": "Example",
+            "match": "!example!"
+        }
+    ]
+}
+```
+
+> In this example, we group "layers" of code ("Component" and "Example" using regex.
```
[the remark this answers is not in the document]
Halleck45: Yes you're right. Using config file is a old (un-documented) feature. I've patched code to use array instead of string for excludes

I've updated the PR with new ewample of configuration

config.json
```diff
@@ -0,0 +1,19 @@
+{
```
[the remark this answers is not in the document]
Halleck45: Just an example (I've renamed it `config-example.json`)

templates/html_report/_header.php
Halleck45: I'm generally in favor if stric equality, but a simple `!` often avoid bugs (here currentGroup is nullable)

src/Hal/Report/Html/Reporter.php
```diff
@@ -23,6 +25,26 @@ class Reporter
      */
     protected $templateDir;
 
+    /**
+     * @var Consolidated[]
+     */
+    private $consolidatedByGroups;
+
+    /**
+     * @var array
```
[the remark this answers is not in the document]
Halleck45: Yes. Fixed in last commit

Halleck45:
Hi @niconoe- ! Nice, thanks for your tests

> It seems that, on the contrary of what your documentation said, the "match" for a group is proceed on the Fully Qualifyed Class Name, and not of the package for instance. You should update the documentation to be more precise about this behavior.

documentation is fixed

> If you want to ensure to match a particular main package in this case, you might want to match something like "\MyPackage". You should explain in the documentation that the match is behaving as a regex, and so you must escape each backslash on 2 levels: 1 for the JSON escape, and the other for the regex escape, Meaning that, according to my example above, you need to write this in your config.json file: "!\\\\MyPackage\\\\!"

documentation is fixed

> A little bug now: the lines of codes, classes, average CC, average bugs in classes (and probably assertions in tests) have indicators based on their previous values. Those previous values inside the groups are not matching the previous values of their respective groups, but the total. To reproduice, simply run twice with the same configuration, and look at any specific group home page.
I don't know the purpose of the "Composer" tab for the groups, maybe it could be removed (if possible)?

Composer and Junit links are removed from sidebar in groups. Trends is removed in page in groups.

> One last thing: as you're including the ability to have a configuration file in this PR, you should list all available options in the documentation and in the example (like the --git or --junit equivalent from CLI).

I've added all missing options as following:

```json
    "plugins": {
        "git": {
            "binary": "git"           // if defined, runs git analyze
        },
        "junit": {
            "report": "/tmp/junit.xml"  // if defined, JUnit report file will be analyzed            
        }
    },
    "extensions": [ "php", "php8" ],
```

> Well, actually, this is even worse: includes option is relative to the location of the configuration file, while report paths are relative to current directory. Please be consistent here, as it's a really strange behavior.

That's not so easy. I think config file will probably versionned with sources, that's why path is 
relative. But report an be generated everywhere on the computer. For example, CI will probably wants generate report from the CI relative directory... I don't know how to be clear with this point.

niconoe-:
Hi @Halleck45 and thank you for that!

For the last topic, you should not assume that config file will be versionned with the sources. You may have single configuration file for a list of similar projects (such as a multi-repo micro-services structure). You can also use Docker to run analysis in a container so any path on the computer can be mounted in a different volume in your container.

So, what you could define as a rule is the same thing that happens in most OS: when the path starts with a "/", it is always absolute, but if not, it is always relative. This way, you allow everything, in an already known behavior, and can be consistent for all options. 

What do you think about?

Halleck45:
@niconoe- I've unified behavior for reports and sources : path are now relative to the config file, or absolutes. Tests are up-to-date.

What do you think of this behavior ?

templates/html_report/css/style.css
Halleck45: I suggest to update CSS later, in another PR

PR #430, opened 2020-06-27
Fix #429, bug when generating html report.

### LINKED ISSUE #429: Bug when generating html report
# Bug report

 

PhpMetrics version: 2.7.0

There is an error when generating HTML report.
````
PHP Notice:  Undefined variable: outDir in /Hal/Report/Html/Reporter.php on line 183
PHP Warning:  file_put_contents(/classes.js): failed to open stream: Permission denied in /src/Hal/Report/Html/Reporter.php on line 184
````

pattisahusiwa:
Lines 181-185 are similar to 139-143. By removing these lines, the bug
will be fixed without affecting the purpose of the function.

### DIFF
--- a/src/Hal/Report/Html/Reporter.php
+++ b/src/Hal/Report/Html/Reporter.php
@@ -178,12 +178,6 @@ class Reporter
             }
         }
 
-        // json data
-        file_put_contents(
-            $outDir . '/classes.js',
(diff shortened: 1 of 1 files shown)


Halleck45:
You're right. A last-minute change caused this regression :( . Thans for your help !

PR #432, opened 2020-06-27
Fix #431, only outdated packages will be reported

### LINKED ISSUE #431: Bug in composer packages report
# Bug report

 

PhpMetrics version: 2.7.0

The packages are in the latest version but reported as outdated

bug

pattisahusiwa:
fix-1

fix-2


Halleck45:
Thanks for the report and the bugfix !

PR #434, opened 2020-06-29
Bump php package required

Mte90:
php5 is not available anymore in debian repository.
Using php-cli let to have the latest php version in the debian system that is php7>= https://packages.debian.org/stretch/php-cli installed on the system.


Halleck45:
Thanks for your help !

PR #440, opened 2020-09-17
fix #436: removing obsolete files

### LINKED ISSUE #436: Debian package is wrong on release page
# Bug report
<s>Looking at the readme the ini format is nmot supported anymore, so I migrated to the json version but also if I pass a broken JSON there are no errors about it.
Using the `help` parameter I saw that `config` parameter is not available but the readme specify it.</s>

Check https://github.com/phpmetrics/PhpMetrics/issues/436#issuecomment-652274498

Mte90:
As #436


Mte90:
ping

Halleck45:
Hi,

these files are here to avoid to broke production CI pipelines using these legacy files (when CI download directly files from github). If we remove them, we'll break ton of pipelines.

I prefer to avoid to remove them. But we can add a README.md and fix the release documentation.

Mte90:
Maybe also add a readme in the folder that explain the purpose. I don't remember how I found those files as a lot of months are passed but I think that I was looking for a tarball or a debian package and found those obsolete files.

Halleck45:
I've just checked, and I confused with the `build` directory. I've just opened #458 to fix it.

Thanks for your help !

PR #446, opened 2020-10-29
Allow usage behind proxy

alediator:
It reads environment variable `http_proxy` before trying to download the composer.json dependency from `Packagist` and creates a stream context to download it if it was configured.

Typically, if you are trying to analyze your code behind a proxy, this tool is stuck getting the packagist dependency tree for all the project dependencies. With this change, its success to download them and move on with the analysis.

### DIFF
--- a/src/Hal/Metric/System/Packages/Composer/Packagist.php
+++ b/src/Hal/Metric/System/Packages/Composer/Packagist.php
@@ -89,13 +89,13 @@ class Packagist
     {
         // Get the environment variable.
         $httpsProxy = getenv('https_proxy');
-        $context    = null;
+        $context = null;
         if ('' !== $httpsProxy) {
             // Create the context.
             $context = stream_context_create(
                 [
                     'http' => [
-                        'proxy' => $this->tcpProtocol($httpsProxy),
+                        'proxy' => str_replace(['http://', 'https://'], 'tcp://', $httpsProxy),
                         'request_fulluri' => true,
                     ],
                 ]
@@ -103,25 +103,5 @@ class Packagist
         }
 
         return json_decode(@file_get_contents($uri, false, $context));
-    }
-
-    /**
-     * Replace http or https protocols with tcp for creating the stream context.
-     *
-     * @param string $httpsProxy
-     *
-     * @return string
-     */
-    private function tcpProtocol($httpsProxy)
-    {
-        return str_replace(
-            'https://',
-            'tcp://',
-            str_replace(
-                'http://',
-                'tcp://',
-                $httpsProxy
-            )
-        );
     }
 }
--- a/src/Hal/Metric/System/Packages/Composer/Packagist.php
+++ b/src/Hal/Metric/System/Packages/Composer/Packagist.php
@@ -36,7 +36,7 @@ class Packagist
         }
         list($user, $name) = explode('/', $package);
         $uri = sprintf('https://packagist.org/packages/%s/%s.json', $user, $name);
-        $json = getURIContentAsJson($uri);
+        $json = $this->getURIContentAsJson($uri);
 
         if (!isset($json->package) || !is_object($json->package)) {
             return $response;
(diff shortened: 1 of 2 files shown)


src/functions.php
```diff
@@ -227,3 +227,32 @@ function percentile($arr, $percentile = 0.95)
...
     return $arr[max(round($percentile * count($arr) - 1.0 - $percentile), 0)];
 }
+
+/**
+ * Download the given URI and decode it as JSON.
+ * @param string $uri
+ *
+ * @return mixed
+ */
+function getURIContentAsJson($uri){
```
niconoe-, line 237:
```suggestion
function getURIContentAsJson($uri)
{
```

src/functions.php
```diff
@@ -227,3 +227,32 @@ function percentile($arr, $percentile = 0.95)
...
+    // get the json file
+    $httpsProxy = getenv('https_proxy');
+    if (!empty($httpsProxy)) {
```
niconoe-, line 240: I'm not a fan of `empty` structure as it type-less comparison.

src/functions.php
```diff
@@ -227,3 +227,32 @@ function percentile($arr, $percentile = 0.95)
...
+        return json_decode(@file_get_contents($uri, false, [
```
niconoe-, line 241: Style is really bad here, but this is not the most important.

Current snippet is incorrect as `file_get_contents`'s 3rd parameter must be a `resource`, not an `array`.
You miss the usage of `stream_context_create` PHP internal function to create the resource based on the parameters you want to set here.

src/functions.php
```diff
@@ -227,3 +227,32 @@ function percentile($arr, $percentile = 0.95)
...
+            'http' => [
+                'proxy'           => str_replace(
```
niconoe-, line 243: Cascading `str_replace` is easily upgradable by using array<string> typed parameters as 1st and 2nd to do several replacements on a single instruction.

src/functions.php
```diff
@@ -227,3 +227,32 @@ function percentile($arr, $percentile = 0.95)
...
+                    str_replace(
+                        'http://',
+                        'tcp://',
+                        $httpsProxy
+                    )
+                ),
+                'request_fulluri' => true,
+            ],
+        ]));
+    } else {
```
niconoe-, line 255: `return` is unskippable in the `if` statement, so there's no need for `else`.

Also, as the `if` condition is negative, I would more inverse the structure and look at it in a positive way. This would improve the readability also as if the env variable is not set, simply return without creating a context, otherwise, (you don't need the `else`) create the context and use it.

src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -36,7 +36,7 @@ public function get($package)
         }
         list($user, $name) = explode('/', $package);
         $uri = sprintf('https://packagist.org/packages/%s/%s.json', $user, $name);
-        $json = json_decode(@file_get_contents($uri));
+        $json = getURIContentAsJson($uri);
```
niconoe-, line 39: I don't know if getting contents from external is something used in another part of the project, but maybe this function could be a method declared in another package to not be in a "messy" library of functions that does all stuff.

niconoe- (changes requested):

alediator:
The test I did was: 

1. `$ export https_proxy=http://host:port`
2. `$ bin/phpmetrics MY_PROJECT_PATH`

With master code it is stuck getting the dependencies whereas with this change it gets the dependencies.

src/Hal/Metric/System/Packages/Composer/Packagist.php
alediator: I've moved it to the class as a private method. I thought it was going to be needed elsewhere, but it wasn't, so it could be there.

src/functions.php
alediator: I think it is solved with the latest push, but let me know if you want me to work it out a bit more.

src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -77,4 +77,51 @@ public function get($package)
...
+    {
+        // Get the environment variable.
+        $httpsProxy = getenv('https_proxy');
+        $context    = null;
+        if ('' !== $httpsProxy) {
+            // Create the context.
+            $context = stream_context_create(
+                [
+                    'http' => [
+                        'proxy' => $this->tcpProtocol($httpsProxy),
```
niconoe-, line 98:
```suggestion
                        'proxy' => str_replace(['http://', 'https://'], 'tcp://', $httpsProxy),
```

And you don't need the `tcpProtocol` method.

src/Hal/Metric/System/Packages/Composer/Packagist.php
```diff
@@ -77,4 +77,51 @@ public function get($package)
...
+     *
+     * @param string $uri
+     *
+     * @return mixed
+     */
+    private function getURIContentAsJson($uri)
+    {
+        // Get the environment variable.
+        $httpsProxy = getenv('https_proxy');
+        $context    = null;
```
niconoe-, line 92:
```suggestion
        $context = null;
```

niconoe- (changes requested):

niconoe- (changes requested):
Some improvements could still be done.

alediator:
Hi again! I would say it's done.

niconoe- (approved):

niconoe-:
Hi @Halleck45 @UFOMelkor !

Like this part is a bit tricky to test, may I ask you your opinion on this?
IMO, I think it's ok as it won't change anything if the `HTTPS_PROXY` env variable is empty, which is the most used case by all of our users I presume (otherwise, we'd get this ticket earlier).

Thanks!

Halleck45:
Hard to test... Thank's for the contribution @alediator :+1:

PR #454, opened 2021-03-26
Removing PHP 7+ return value typehint

crimsonkissaki:
This package claims to be good for PHP >=5.5 so it cannot include PHP 7 only syntax.

$ phpmetrics --report-html=tests/_reports/metrics/report.html ./src
PHP Parse error:  syntax error, unexpected ':', expecting '{' in C:\Users\ejohnson8\Zend\workspaces\stdlib\vendor\phpmetrics\phpmetrics\src\Hal\Application\Config\Validator.php on line 51

### DIFF
--- a/src/Hal/Application/Config/Validator.php
+++ b/src/Hal/Application/Config/Validator.php
@@ -48,7 +48,7 @@ class Validator
         }
         $groupsRaw = $config->get('groups');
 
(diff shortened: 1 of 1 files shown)


Halleck45:
Yes, you're right. Thanks !

Issue #451, opened 2021-02-11
use of bubble graph on overview page is hampered by animation


MarkProDev:
# Bug report

 

This is an easy one.  It would be nice to use the overview to see problem modules graphically but the animation moves the graph making it difficult to identify the problem 'children'.  Either disable the animation if a user clicks on one of the images or mouses over the image or offer another view, i.e. side by side.  I am using version 2.7.4

Halleck45:
Hi,

I'm agree. What do you think about https://github.com/phpmetrics/PhpMetrics/pull/455 ?

Now you must click to toggle between these charts

![Capture d’écran du 2021-06-06 22-46-09](https://user-images.githubusercontent.com/1076296/120939595-0d871b00-c719-11eb-98e6-cf4570386e25.png)

MarkProDev:
sounds great!  thank you!!!

PR #455, opened 2021-06-06
[ISSUE-451] remove auto scrolling of charts

Halleck45:
Fixes https://github.com/phpmetrics/PhpMetrics/issues/451


templates/html_report/js/graph-maintainability.js
```diff
@@ -1,14 +1,23 @@
-function chartMaintainability(withoutComment)
-{
-    var chartId = withoutComment ? 'svg-maintainability-without-comments' : 'svg-maintainability';
-
+function chartMaintainability(withoutComment) {
+    var chartId = 'svg-maintainability';
+    withoutComment = typeof (withoutComment) !== 'undefined' ? withoutComment : false;
```
niconoe-, line 3: I think you should either:
- force the default graph to be with comments and so, remove the ability to call the method `chartMaintainability` with an argument; or
- let the graph to be loaded being decided by the UI when the call is made, and keep the argument, but in this case, force the call to be with `true` or `false`.

Therefore, in any case, this default value setting for `withoutComment` should be removed.

templates/html_report/_footer.php
```diff
@@ -183,7 +183,6 @@
 <script type="text/javascript" src="<?php echo $this->assetPath; ?>js/d3.hexbin.v0.js"></script>
 <script type="text/javascript" src="<?php echo $this->assetPath; ?>js/sort-table.min.js"></script>
 <script type="text/javascript" src="<?php echo $this->assetPath; ?>js/graph-maintainability.js"></script>
-<script type="text/javascript" src="<?php echo $this->assetPath; ?>js/graph-carousel.js"></script>
```
niconoe-, line 186: This file is not deleted in the MR, is there any other uses?

niconoe- (changes requested):

templates/html_report/js/graph-maintainability.js
Halleck45: Hi,

sorry, I don't understand

> force the default graph to be with comments and so, remove the ability to call the method chartMaintainability with an argument; or
Now by default the graph is "with comments". But I need to pass an argument to change behavior on switch (via the `toggleChartMaintainability` function)

> let the graph to be loaded being decided by the UI when the call is made, and keep the argument, but in this case, force the call to be with true or false.

In this case it will be difficult to switch easily the kinf of chart by click.

I think I misunderstood your comment. Do you have any code suggestion to help me to understand ?

templates/html_report/index.php
```diff
@@ -91,22 +91,17 @@
...
-                                        <div class="bloc-graph-item first">
-                                            <div class="label">Maintainability / complexity</div>
-                                            <div id="svg-maintainability" class="svg-container"></div>
-                                        </div>
-                                        <div class="bloc-graph-item second">
-                                            <div class="label">Maintainability without comments / complexity</div>
-                                            <div id="svg-maintainability-without-comments" class="svg-container"></div>
-                                        </div>
+                            <div>
+                                <div>
```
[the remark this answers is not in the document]
Halleck45: Fixed

templates/html_report/_footer.php
Halleck45: Yes, you're right. That's fixed

templates/html_report/js/graph-maintainability.js
niconoe-: Ok, I misread something in your code and I didn't noticed that `chartMaintainability` is called with an argument within `toggleChartMaintainability`.

So, my suggestion here is to call `chartMaintainability(false);` (give the argument) to force the `chartMaintainability` function to recieve an argument. It's clearer than defining a default value based on the fact the variable is defined or undefined:

```suggestion
```

And, both usages of `chartMaintainability();` in `templates/html_report/index.php` and `templates/html_report/panel.php` must be replaced by `chartMaintainability(false)`.

Halleck45: Hi,

I suggest to merge this for the moment. I'll try to improve it after

Issue #441, opened 2020-10-09
JUnit config parameter docs don't match what the code actually does


AlexHowansky:
# Bug report

## Version 2.7.4

The `README.md` file includes this configuration file example of linking to a JUnit report:

```JSON
"junit": {
    "file": "/tmp/junit.xml"
}
```

However, the code in `Hal\Application\Config\File\ConfigFileReaderJson` uses the key `report` and not the key `file`:

```PHP
if (isset($jsonData['plugins'], $jsonData['plugins']['junit'], $jsonData['plugins']['junit']["report"])) {
```

Thus, either the `README.md` should be changed to:

```JSON
"junit": {
    "report": "/tmp/junit.xml"
}
```

Or the `Hal\Application\Config\File\ConfigFileReaderJson` should be changed to:

```PHP
if (isset($jsonData['plugins'], $jsonData['plugins']['junit'], $jsonData['plugins']['junit']["file"])) {
```

I did not create a pull request for this change as I don't know which of these two keys was intended and it's trivial enough to implement.

Halleck45:
Thanks for the report. This will be fixed with https://github.com/phpmetrics/PhpMetrics/pull/457

PR #457, opened 2021-06-07
[ISSUE-441] Fix junit parameter in json config file

Halleck45:
Fixes https://github.com/phpmetrics/PhpMetrics/issues/441

### DIFF
--- a/src/Hal/Application/Config/File/ConfigFileReaderJson.php
+++ b/src/Hal/Application/Config/File/ConfigFileReaderJson.php
@@ -122,16 +122,4 @@ class ConfigFileReaderJson implements ConfigFileReaderInterface
             }
         }
     }
-
-    /**
-     * @return string
-     */
-    private function resolvePath($path)
-    {
-        if (DIRECTORY_SEPARATOR !== $path[0]) {
-            $path = dirname($this->filename) . DIRECTORY_SEPARATOR . $path;
-        }
-
-        return $path;
-    }
 }
--- a/src/Hal/Application/Config/File/ConfigFileReaderJson.php
+++ b/src/Hal/Application/Config/File/ConfigFileReaderJson.php
@@ -99,19 +99,19 @@ class ConfigFileReaderJson implements ConfigFileReaderInterface
             }
         }
 
-        if (isset($jsonData['plugins'], $jsonData['plugins']['git'], $jsonData['plugins']['git']['binary'])) {
+        if (isset($jsonData['plugins']['git']['binary'])) {
             $config->set('git', $jsonData['plugins']['git']['binary']);
         }
 
-        // retrocompatibility with typo in documentation
+        // backward compatibility with typo in documentation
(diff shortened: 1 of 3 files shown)


src/Hal/Application/Config/File/ConfigFileReaderJson.php
```diff
@@ -72,6 +72,13 @@ public function read(Config $config)
             $config->set('git', $jsonData['plugins']['git']['binary']);
         }
 
+        // retrocompatibility with typo in documentation
+        // see https://github.com/phpmetrics/PhpMetrics/issues/441
+        // file -> report
+        if (isset($jsonData['plugins'], $jsonData['plugins']['junit'], $jsonData['plugins']['junit']["file"])) {
```
niconoe-, line 78: As `isset` is already dealing with possible unset variable, you can directly use `$jsonData['plugins']['junit']["file"]`

```suggestion
        if (isset($jsonData['plugins']['junit']['file'])) {
```

src/Hal/Application/Config/File/ConfigFileReaderJson.php
```diff
@@ -72,6 +72,13 @@ public function read(Config $config)
...
+            $jsonData['plugins']['junit']["report"] = $jsonData['plugins']['junit']["file"];
```
niconoe-, line 79:
```suggestion
            $jsonData['plugins']['junit']['report'] = $jsonData['plugins']['junit']['file'];
```

src/Hal/Application/Config/File/ConfigFileReaderJson.php
```diff
@@ -72,6 +72,13 @@ public function read(Config $config)
...
+        }
+
         if (isset($jsonData['plugins'], $jsonData['plugins']['junit'], $jsonData['plugins']['junit']["report"])) {
```
niconoe-, line 82: Same remarks as for `'file` key, you can reduce the size of your `isset` and replace double quotes with simple quotes for homogeneity.

Halleck45: Hi,

isset throw warning on old PHP versions. We need to check each depth of array to avoid warnings.

niconoe-: Tested with 3v4l.org through PHP 4.3.0 to PHP 8.0.3 and no warning shows: https://3v4l.org/MYRHv

This is not a configuration that does not display it, because warnings/notices are displayed in the case we are using the variable: https://3v4l.org/U2rnH

If you can experiment a warning without checking all depth, I'm curious to see how and do some tests about this.

niconoe-:
Hi @Halleck45 

I did the changes I requested. It should be OK now. Do you want to give a last check before I merge the PR?

Halleck45:
Hi !

@niconoe- can you rebase please ?

PR #464, opened 2021-09-19
Fixed issue with incorrect table header on the complexity report

denniscoorn-paqt:
PR #421 accidentally moved the column "Relative system complexity" to the end of the table header causing an issue where the order of the header columns don't match the order of the data columns. A unit test has been added to validate the expected order of the columns.

Before:
Screenshot 2021-09-19 at 17 12 30

After:
Screenshot 2021-09-19 at 17 09 30

Note that in the second image the data values for "Bugs" and "Defects" are correctly displayed in their corresponding column.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
     "require-dev": {
         "phpunit/phpunit": "^4.8.36 || ^5.7.27 || ^6.5.14",
         "sebastian/comparator": ">=1.2.3",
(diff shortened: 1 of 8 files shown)


niconoe- (approved):
Ty for your PR!

Looks good to me.

Halleck45:
Thanks @denniscoorn-paqt 

Adding tests on HTML report is a great idea :clap:

PR #466, opened 2021-11-25
Remove php8.1 deprecations

niconoe-:
After some tests with PHP 8.1, I have some deprecation errors, mainly due to [this RFC](https://wiki.php.net/rfc/internal_method_return_types), but also because of a stricter type check internally done by PHP now.

There are 3 commits for 3 different kind of resolutions depending on the issues.

1. I removed the usage of the "flag" filter for the Finder, as it was never set, so always `null`, while an integer is expected.
2. I added the PHP Attribute `#[ReturnTypeWillChange]` on all methods that implement a PHP internal method. This is because they now need to have the expected return type defined, but as PhpMetrics is still compatible with PHP >= 5.5 and return type definitions are PHP >= 7, we can't provide the expected fix. We need to drop the support of PHP 5.5 and PHP 5.6 minimum to fix this correctly (and it will be necessary for PHP 9). While PHP Attributes are PHP 8 only, their syntax allows their usages in PHP < 8 as they'll just be parsed as comments, so there's no side effects. On PHP 8.0, this attribute has no effect neither.
3. I added the "unknown" status along side "outdated" or "latest" for composer packages that are private. As we can't get more information on them, it would caused "installed" and "latest" values set to `null`, which broke the usage of `version_compare` to set the right status on the package. This new "unknown" status has been added to the UI too, and it declares the package as a private package, with a light-blue backgound, on the opposite of the orange background when a package is outdated.

If you have any remarks, please tell me. I'll see how this could be improved ASAP as PHP 8.1 is being released today.

### DIFF
--- a/src/Hal/Component/File/Finder.php
+++ b/src/Hal/Component/File/Finder.php
@@ -39,22 +39,13 @@ class Finder
     private $excludedDirs = [];
 
     /**
-     * Flags for RecursiveDirectoryIterator
-     *
-     * @var integer
-     */
-    private $flags;
-
-    /**
      * @param string[] $extensions   regex of file extensions to include
      * @param string[] $excludedDirs regex of directories to exclude
-     * @param int $flags
      */
-    public function __construct(array $extensions = ['php'], array $excludedDirs = [], $flags = null)
+    public function __construct(array $extensions = ['php'], array $excludedDirs = [])
     {
         $this->extensions = $extensions;
         $this->excludedDirs = $excludedDirs;
-        $this->flags = $flags;
     }
 
     /**
@@ -69,7 +60,7 @@ class Finder
         foreach ($paths as $path) {
             if (is_dir($path)) {
                 $path = rtrim($path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR;
-                $directory = new RecursiveDirectoryIterator($path, $this->flags);
+                $directory = new RecursiveDirectoryIterator($path);
                 $iterator = new RecursiveIteratorIterator($directory);
 
                 $filterRegex = sprintf(
--- a/src/Hal/Component/Tree/Graph.php
+++ b/src/Hal/Component/Tree/Graph.php
@@ -100,6 +100,7 @@ class Graph implements \Countable
     /**
      * @return int
      */
+    #[\ReturnTypeWillChange]
     public function count()
     {
         return count($this->data);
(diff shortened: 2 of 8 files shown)


designermonkey:
@niconoe- does this actually work under php8.1 for you? While the code is now type safe and doesn't error, I get no results when running it under php8.1

I wish there were tests so I could narrow it down.

Here's my output:

```
bin/phpmetrics -c phpmetrics.json


Executing system analyzes...

Executing composer analyzes, requesting https://packagist.org...

LOC
    Lines of code                               0
    Logical lines of code                       0
    Comment lines of code                       0
    Average volume                              0
    Average comment weight                      0
    Average intelligent content                 0
    Logical lines of code by class              0
    Logical lines of code by method             0

Object oriented programming
    Classes                                     0
    Interface                                   0
    Methods                                     0
    Methods by class                            0
    Lack of cohesion of methods                 0
    
Coupling
    Average afferent coupling                   0
    Average efferent coupling                   0
    Average instability                         0
    Depth of Inheritance Tree                   0
    
Package
    Packages                                    0
    Average classes per package                 0
    Average distance                            0
    Average incoming class dependencies         0
    Average outgoing class dependencies         0
    Average incoming package dependencies       0
    Average outgoing package dependencies       0

Complexity
    Average Cyclomatic complexity by class      0
    Average Weighted method count by class      0
    Average Relative system complexity          0
    Average Difficulty                          0
    
Bugs
    Average bugs by class                       0
    Average defects by class (Kan)              0

Violations
    Critical                                    0
    Error                                       0
    Warning                                     0
    Information                                 0




Done
```

designermonkey:
@niconoe- I've spotted a missed one:

```
PHP Deprecated:  abs(): Passing null to parameter #1 ($num) of type int|float is deprecated in /Users/john/Projects/jorpo-co/concerto/concerto-api/libs/phpmetrics/phpmetrics/src/Hal/Violation/Package/StableAbstractionsPrinciple.php on line 24
```

designermonkey:
Also, I can't get this to provide any output other than the CLI on php8.1

Is it working for you?

caugner:
Fixes #468.

caugner:
> I've spotted a missed one:

The explanation seems to be that `PackageMetric::getDistance()` may return `null`.

To keep the current behaviour and avoid that warning, `StableAbstractionsPrinciple::apply()` would need to ignore metrics without any distance.

PR #467, opened 2021-12-01
Add command to get summary as Json

roelbaka:
I wanted to add the summary metrics of PHPMetrics to a build pipeline so I can send this data to a monitoring client to track changes over time.

 I did not find an easy way to do this, so I thought I would add a functionality to get the consolidated data as JSON.

image


Halleck45:
Hi @roelbaka ,

thanks for your contribution ! That's a good idea.

PR #472, opened 2022-03-22
Way to search patterns of code

Halleck45:
Hi,

* This PR **allows to search patterns of code**. Instead of providing our own rules, now we can add dynamic rules for violations.

```
searches:
    Repository which uses Service:
        type: class
        instanceOf:
            - App\MyRepository
        nameMatches: ".*Repository.*"
        usesClasses:
            - ".*Service"
        failIfFound: true
        
    Class with too complex code:
        type: class
        ccn: ">=10"
        failIfFound: true
```

This PR also:
-----

* allows also to use **yaml for configuration.**
* **adds a `--metrics` option**
* **excludes getters and setters** for ccn and lcom calculation
* adds a **composer option**


gmponos (approved):

niconoe-:
This looks like awesome!

If I understand well, then there would be no need to keep the current defined violations principles (unless for history) and ue a default configuration that include the related searches, right?

niconoe- (approved):

Halleck45:
> If I understand well, then there would be no need to keep the current defined violations principles (unless for history) and ue a default configuration that include the related searches, right?

Exactly :smile:

Issue #444, opened 2020-10-26
Show cyclomatic complexity of methods


Mte90:
# Feature request

Right now the tool show the total of the class of cyclomatic complexity but maybe there are some methods that require more than the other.

kudashevs:
The PR #481 adds the possibility to see the cyclomatic complexity of all of the methods in file.

niconoe-:
Based on @kudashevs's work, this is now available in v3.0.0rc4. Closing.

PR #481, opened 2022-11-06
Add a CCN value to every method in a class

kudashevs:
I've found your package very useful. I want to thank you for the great package.

During the usage, I realized that there is a thing which I really miss in this package. The possibility to see the Cyclomatic complexity of the most complex method in a class is useful, but I really wanted to see the complexity of every method in a class. So, my proposition is to add this possibility, which I think is quite useful. For now it looks like this.

image

### DIFF
--- a/templates/html_report/css/style.css
+++ b/templates/html_report/css/style.css
@@ -443,6 +443,29 @@ a[target="_blank"]::before {
     color: #333;
     font-size: 0.9em;
 }
+.details {
+    display:none;
+    position:absolute;
+    width:100%;
+    margin:30px 0 0;
+    left:0;
+    right:0;
+    z-index:999;
+    background:#fff;
+}
+.table {
+    display:table;
+    table-layout:fixed;
+    width:100%;
+}
+.methods-list {
+    display:table-row;
+    text-align:left;
+}
+.methods-list > span {
+    display:table-cell;
+    padding:.125em 0 .125em 0;
+}
 
 /* -------- charts ---------------- */
 .tooltip {
@@ -548,7 +571,15 @@ table tr td {
     padding: 0 .5em;
     cursor: pointer;
 }
-
+.js-sort-table {
+    position:relative;
+}
+.js-sort-table tr {
+    position:relative;
+}
+.js-sort-table td {
+    padding:.125em .325em .125em 0;
+}
 .js-sort-table thead th {
     cursor: pointer;
 }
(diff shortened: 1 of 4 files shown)


niconoe-:
Thank you very much for this feature, it's an interesting one to go deeper into the details!

On UI/UX side, I'm afraid this could be too verbose on projects with many classes. Do you think you could handle it via an accordion on the class line? A kind of triange-arrow-right could be added for each class-line and onclick on that arrow or the class name, the accordion would collapse (triangle-arrow rotates to be oriented to bottom) revealing the CCN for every methods of the current clicked class.

The default verbosity would still be the same so that current users won't get lost by too much data displayed, but at least the details by method would become available on demand thanks to your work.

niconoe-:
Thank you very much! This looks nice to me!

@Halleck45 what do you think about?

niconoe- (approved):

templates/html_report/css/style.css
```diff
@@ -578,7 +578,7 @@ table tr td {
     position:relative;
 }
 .js-sort-table td {
-    padding:.125em .325em .125em 0;
+    vertical-align:baseline;
```
kudashevs, line 581: This one removes an element jumping which happens when you click on a class name.

Mte90:
Maybe for next 3.0.0 @niconoe-?

kudashevs:
Hello @Halleck45 

Could you please provide any feedback about this PR?

This functionality was requested long before in #444 and I find it very useful.
Do you like the whole idea? Should I refine or improve something in this PR?

niconoe-:
I think that rather having a table component, we should migrate to something more flexible to allow sub-tables. I'll try to take some time to search for reliable components about it.

niconoe-:
Sorry for the waiting about the PR. I had to take a long break recently.

@kudashevs  I reworked the incoming v3 with your work in an adapted way in order to manage both UX/UI and not to interfear too much with logic.

I go further than your proposal by managing to display the accessors even if they are ignored from the cyclomatic complexity calculations, with a flag to explain this to the end-user.
I also managed to open the rows via an accordion style without having too much JS changed, and keeping the possibilty to sort rows by their column values.

Here is a sample of how it's displayed now:
selection

I think this way, it's better, and it opens the way to having other metrics on the methods in the future.

It will be part of the v3 and will probably be released in the next RC.

Feel free to suggest or remark stuff about it 😉

kudashevs:
It looks good. Hope to see this functionality soon in the V3 RC.

PR #497, opened 2023-09-13
SystemComplexityVisitor for methods with the same name on different objects

andrii-pukhalevych:
`(new B)->bar()` and `(new C)->bar()` should be calculated separately.

### DIFF
--- a/src/Hal/Metric/Class_/Structural/SystemComplexityVisitor.php
+++ b/src/Hal/Metric/Class_/Structural/SystemComplexityVisitor.php
@@ -50,14 +50,26 @@ class SystemComplexityVisitor extends NodeVisitorAbstract
                     $output = 0;
                     $fanout = [];
 
-                    iterate_over_node($node, function ($node) use (&$output, &$fanout) {
+                    $parentNode = $node;
+                    iterate_over_node($node, function ($node) use (&$output, &$fanout, $parentNode) {
                         switch (true) {
                             case $node instanceof Stmt\Return_:
                                 $output++;
                                 break;
                             case $node instanceof Node\Expr\StaticCall:
+                                $class = getNameOfNode($node->class);
+                                if ('static' === $class || 'self' === $class) {
+                                    $class = getNameOfNode($parentNode);
+                                }
+                                $fanout[] = $class . '::' . getNameOfNode($node->name) . '()';
+                                break;
                             case $node instanceof Node\Expr\MethodCall:
-                                array_push($fanout, getNameOfNode($node));
+                                $class = getNameOfNode($node->var);
+                                if ('this' === $class) {
+                                    $class = getNameOfNode($parentNode);
+                                }
+                                $fanout[] = $class . '->' . getNameOfNode($node->name) . '()';
+                                break;
                         }
                     });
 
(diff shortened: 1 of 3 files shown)


niconoe-:
Thank you for your contribution. However, I'm unsure what kind of issue are you solving here. Could you please elaborate and give an example?

Thanks again.

andrii-pukhalevych:
https://github.com/phpmetrics/PhpMetrics/blob/4b77140a11452e63c7a9b98e0648320bf6710090/src/Hal/Metric/Class_/Structural/SystemComplexityVisitor.php#L58C1-L65C1

In current version, `(new B)->bar()` and `(new C)->bar()` counted as the same method. Because `getNameOfNode($node)` returns `'B'` for both calls, and after `array_unique`,  `$fanout = 1`.
Since its 2 different methods, it should be counted as 2 separate methods.

andrii-pukhalevych:
Example:
```
class A {

    public function foo($a, $b, $c)
    {
        (new C)->bar();
        (new B)->bar();
    }
}

class B {
    public function bar()
    {

    }
}

class C {
    public function bar()
    {

    }
}
```

niconoe-:
Ok, fine, I got it now! Thanks a lot.
I'll integrate this into the new rc version.

andrii-pukhalevych:
Agree, [for this test file I updated](https://github.com/andrii-pukhalevych/PhpMetrics/blob/d1716a96c863dbba2d4e94a65a9e9baecdfa358b/tests/Metric/examples/systemcomplexity1.php), 3.x-dev branch results are almost the same as results from PR I did.

For class `A`:
andrii-pukhalevych:patch1 : `{DataComplexity: 0.5, StructuralComplexity: 36.0, SystemComplexity: 36.5]`
3.x-dev: `{DataComplexity: 0.25, StructuralComplexity: 40.5, SystemComplexity: 40.75}`

Not sure what is correct, but the difference is quite small.

Anyway, please merge this PR into version 2, because many PHP 7 users still will use it few years :)
Or I can close it, if it is not possible to make version 2 release.

niconoe-:
> Agree, [for this test file I updated](https://github.com/andrii-pukhalevych/PhpMetrics/blob/d1716a96c863dbba2d4e94a65a9e9baecdfa358b/tests/Metric/examples/systemcomplexity1.php), 3.x-dev branch results are almost the same as results from PR I did.
> 
> For class `A`: andrii-pukhalevych:patch1 : `{DataComplexity: 0.5, StructuralComplexity: 36.0, SystemComplexity: 36.5]` 3.x-dev: `{DataComplexity: 0.25, StructuralComplexity: 40.5, SystemComplexity: 40.75}`
> 
> Not sure what is correct, but the difference is quite small.

The 3.x-dev branch comes with more elements to analyse and compare, so it might explain the difference. In any kind, you're right, the differences are small enough agree this is fixed on v3

> Anyway, please merge this PR into version 2, because many PHP 7 users still will use it few years :) Or I can close it, if it is not possible to make version 2 release.

I'm sorry but even if I merge this PR in v2, I wouldn't be able to release a patch version. The releasing system has been reviewed in v3 in order to make it possible for other people than Jean-François to release, but on v2 that's not the case, unfortunately.
That's the main reason I won't support PhpMetrics v2 anymore.

andrii-pukhalevych:
I would like this PR to be merged in v2, even without release. Maybe it will help somebody. 
If you dont want, then just close, NP :)

niconoe-:
Ok then :slightly_smiling_face:

PR #509, opened 2024-08-06
Fix packagist array key type error

Klemo1997:
- if there is an integer string set as a key of associative array, it will automatically be converted to integer, which causes the call of `ltrim` to fail, because of `declare(strict_types=1)` configuration
- this caused a runtime failure in my project


niconoe-:
Thanks!

Klemo1997:
You're welcome! BTW this tool is beyond awesome!

PR #513, opened 2025-01-05
remove unnecessary whitespaces

YAhiru:


niconoe-:
Thank you for fixing this, but current `master` branch is not aligned with the released v3 (still in RC). I strongly encourage you to use the latest tag.

PR #523, opened 2025-04-09
Support on nikic/php-parser v5.x

AlekVolsk:
This is required for compatibility with other libraries to ensure that PHP 8.4 works.


niconoe-:
Thank you for your support!

PhpMetrics master currently refers to the latest 2.x version while a 3.x branch is available and support php-parser in v5

If you need to use PhpMetrics with this parser, please use the latest 3.x release, even if still tagged in rc.

PR #531, opened 2025-10-02
fix trends not displayed correctly

zimmer-yan:
PhpMetrics v2.9.1

Description:
Fixing trends not being displayed correctly.

Current behavior:
The trends in the html report will always show the current value and a trend of minus that value.

Correct behavior:
show actual trend change

Root cause:
`Hal\Report\Html\Reporter.php:255` checks for `isset($this->sum->$type->$key)` which will never be set, resulting in `$newValue` being assigned 0, causing the trend to be `0 - $oldValue`.

Fix:
Changing the lookup to `$this->$type->$key` from `$this->sum->$type->$key` checks for the correct variable.

### DIFF
--- a/src/Hal/Report/Html/Reporter.php
+++ b/src/Hal/Report/Html/Reporter.php
@@ -252,7 +252,7 @@ class Reporter
         }
 
         $oldValue = $last->$type->$key;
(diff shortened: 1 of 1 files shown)


BernardNganduDev (approved):

Halleck45:
Well done @zimmer-yan , thanks

PR #535, opened 2026-07-08
Import PhpParser Error class in PrepareParserTask

andrii-pukhalevych:
Fix that single file with a syntax error aborts the whole analysis and no report is generated. PhpParser\Error extends RuntimeException (i.e. it is an \Exception, not an \Error), so the catch clause never matches and the exception bubbles up, killing the run.

Steps to reproduce:
Create a project containing one valid PHP file and one file with invalid PHP. Real-world case — doptor/Doptor ships module templates with ***ROUTES*** placeholders inside .php files.


niconoe-:
Thanks!

PR #536, opened 2026-07-16
Fix PHP 8.2+ deprecation about dynamic property

davidmpaz:
Hi,

thanks for this project! While using it in some other project of mine with PHP 8.5 I found this deprecation. I thought it would be nice to have it in place to avoid unnecessary warning when using the tool. After all it makes nothing, just declaring the properties.

----

For PHP 8.2+ the usage of dynamics properties are already deprecated. This commit add the properties to the class declaration.


Halleck45:
Hi @davidmpaz ,

thanks for the contribution ! :clap:

PR #537, opened 2026-07-16
Fix tests warning and notices

davidmpaz:
Hi,

thanks for maintaining this project. Could we include this on the main-stream, these are some details I noticed when running the tests for the project when creating  https://github.com/phpmetrics/PhpMetrics/pull/536

----

Add returns types to HashMap interface. 
Correct namespace on IssuerTest. 
Avoid using warning silence operator.

### DIFF
--- a/src/Hal/Component/Tree/HashMap.php
+++ b/src/Hal/Component/Tree/HashMap.php
@@ -20,7 +20,7 @@ class HashMap implements \Countable, \IteratorAggregate
      * @param Node $node
      * @return $this
      */
-    public function attach(Node $node)
+    public function attach(Node $node): self
     {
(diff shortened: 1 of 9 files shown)


Halleck45:
Hi,

I cannot accept this PR : it breaks retro-compatibility.

It's sometime hard, but it's probably possible to keep the compatibility with all PHP versions, even in this case. 

Edit : I've just merged https://github.com/phpmetrics/PhpMetrics/pull/538 to fix the depreciations on CI

davidmpaz:
Hi, 

sure totally understandable. Let me check on this again. Maybe there is a sweet spot for reducing some warning while keeping fully backward compatible

Thanks
David

